PR #1, opened 2016-11-29
Implement group backend

PVince81:
Implement group manager.

To test:

1. Enable app
1. Create three users "user1", "user2", "user3"
1. `insert into oc_custom_group values (NULL, 'myfriends', 'My friends');`
1. `insert into oc_custom_group_member values (1, 'user2', 1);`
1. `insert into oc_custom_group_member values (1, 'user3', 0);`
1. Login as "user1"
1. Create a folder "test"
1. Share with "My fri" let it autocomplete: select "customgroup_1" (need future work to support display names)
1. Login as "user2" and see that the folder is visible

Some concerns:
- current impl lets any user share with anybody, unless admin ticks in admin page "only allow sharing with groups in which user is member", this could be fine as is...
- need display name support in core, might need to change some group manager APIs to support returning arrays instead of just the group ids

@DeepDiver1975 @butonic @jvillafanez what do you think ?

### DIFF
--- a/lib/Application.php
+++ b/lib/Application.php
@@ -2,7 +2,7 @@
 /**
  * @author Vincent Petry <pvince81@owncloud.com>
  *
- * @copyright Copyright (c) 2016, ownCloud GmbH.
+ * @copyright Copyright (c) 2016, ownCloud, Inc.
  * @license AGPL-3.0
  *
  * This code is free software: you can redistribute it and/or modify
@@ -19,5 +19,20 @@
  *
  */
 
-$app = new \OCA\CustomGroups\Application();
-$app->registerGroupBackend();
+namespace OCA\CustomGroups;
+
+use OCP\AppFramework\App;
+
+class Application extends App {
+	public function __construct (array $urlParams = array()) {
+		parent::__construct('customgroups', $urlParams);
+	}
+
+	/**
+	 * Register the group manager
+	 */
+	public function registerGroupBackend() {
+		$backend = $this->getContainer()->query('\OCA\CustomGroups\CustomGroupsBackend');
+		$this->getContainer()->getServer()->getGroupManager()->addBackend($backend);
+	}
+}
--- a/lib/CustomGroupsDatabaseHandler.php
+++ b/lib/CustomGroupsDatabaseHandler.php
@@ -1,0 +1,356 @@
+<?php
+/**
+ * @author Vincent Petry <pvince81@owncloud.com>
+ *
+ * @copyright Copyright (c) 2016, ownCloud GmbH.
+ * @license AGPL-3.0
+ *
+ * This code is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License, version 3,
+ * as published by the Free Software Foundation.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License, version 3,
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>
+ *
+ */
+
+namespace OCA\CustomGroups;
+
+use OCP\IDBConnection;
+use OCP\ILogger;
+
+/**
+ * Database handler for custom groups
+ */
+class CustomGroupsDatabaseHandler {
+
+	/**
+	 * @var IDBConnection
+	 */ 
+	private $dbConn;
+
+	/**
+	 * @var ILogger
+	 */ 
+	private $logger;
+
+
+	public function __construct(IDBConnection $dbConn, ILogger $logger) {
+		$this->dbConn = $dbConn;
+		$this->logger = $logger;
+	}
+
+	/**
+	 * Checks whether the user is member of a group or not.
+	 *
+	 * @param string $uid uid of the user
+	 * @param int $numericGroupId id of the group
+	 * @return bool
+	 * @throws \Doctrine\DBAL\Exception\DriverException in case of database exception
+	 */
+	public function inGroup($uid, $numericGroupId) {
+		$qb = $this->dbConn->getQueryBuilder();
+
+		$cursor = $qb->select('user_id')
+			->from('custom_group_member')
+			->where($qb->expr()->eq('group_id', $qb->createNamedParameter($numericGroupId)))
+			->andWhere($qb->expr()->eq('user_id', $qb->createNamedParameter($uid)))
+			->execute();
+
+		$result = $cursor->fetch();
+		$cursor->closeCursor();
+
+		return $result ? true : false;
+	}
+
+	/**
+	 * Get all groups a user belongs to
+	 *
+	 * @param string $uid Name of the user
+	 * @return array an array of numeric group ids
+	 * @throws \Doctrine\DBAL\Exception\DriverException in case of database exception
+	 */
+	public function getUserGroups($uid) {
+		$qb = $this->dbConn->getQueryBuilder();
+		$cursor = $qb->select('group_id')
+			->from('custom_group_member')
+			->where($qb->expr()->eq('user_id', $qb->createNamedParameter($uid)))
+			->orderBy('group_id', 'ASC')
+			->execute();
+
+		$groups = [];
+		while ($row = $cursor->fetch()) {
+			$groups[] = (int)$row['group_id'];
+		}
+		$cursor->closeCursor();
+
+		return $groups;
+	}
+
+	/**
+	 * Searches groups by display names where the given
+	 * search string can appear anywhere within the display name
+	 * field.
+	 *
+	 * @param string $search search string
+	 * @param int $limit limit or -1 to disable
+	 * @param int $offset offset
+	 * @return array an array of group info
+	 * @throws \Doctrine\DBAL\Exception\DriverException in case of database exception
+	 */
+	public function searchGroups($search = '', $limit = -1, $offset = 0) {
+		$qb = $this->dbConn->getQueryBuilder();
+		$qb->select(['group_id', 'uri', 'display_name'])
+			->from('custom_group')
+			->orderBy('display_name', 'ASC');
+
+		if ($search !== '') {
+			$likeString = '%' . $this->dbConn->escapeLikeParameter(strtolower($search)) . '%';
+			$qb->where($qb->expr()->like($qb->createFunction('LOWER(`display_name`)'), $qb->createNamedParameter($likeString)));
+		}
+
+		if ($limit > 0) {
+			$qb->setMaxResults($limit);
+		}
+
+		if ($offset !== 0) {
+			$qb->setFirstResult($offset);
+		}
+
+		$cursor = $qb->execute();
+		$groups = $cursor->fetchAll();
+		$cursor->closeCursor();
+		return $groups;
+	}
+
+	/**
+	 * Returns the info for a given group.
+	 *
+	 * @param string $numericGroupId group id
+	 * @return array|null group info or null if not found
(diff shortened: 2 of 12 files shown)


PVince81:
also:

- custom groups are visible in user manager. Can't hide them because making `getGroups` return an empty would break searching in the sharing autocomplete. Might need changing core APIs to let the manager specify if listable or not.

lib/CustomGroupsManager.php
```diff
@@ -0,0 +1,196 @@
...
+
+namespace OCA\CustomGroups;
+
+/**
+ * Group manager for custom groups.
+ */
+class CustomGroupsManager implements \OCP\GroupInterface {
+
+	const GROUP_ID_PREFIX = 'customgroup_';
+	const GROUP_ID_PREFIX_LEN = 12;
```
jvillafanez: I don't think it's a good idea to hardcode this. The extra cost of calculating the string's length should be meaningless, and we would only maintain one constant.

lib/CustomGroupsManager.php
```diff
@@ -0,0 +1,196 @@
...
+	 * @return array an array of group names
+	 *
+	 */
+	public function getGroups($search = '', $limit = -1, $offset = 0) {
+		$parameters = [];
+		$searchLike = '';
+		// FIXME: search by display name instead
+		if ($search !== '') {
+			$parameters[] = '%' . $search . '%';
+			$searchLike = ' WHERE LOWER(`displayname`) LIKE LOWER(?)';
```
jvillafanez: Do we want lowercase comparations?

lib/CustomGroupsManager.php
```diff
@@ -0,0 +1,196 @@
...
+
+	/**
+	 * Checks whether the user is member of a group or not.
+	 *
+	 * @param string $uid uid of the user
+	 * @param string $gid gid of the group
+	 * @return bool
+	 */
+	public function inGroup($uid, $gid) {
+		$numericGroupId = $this->extractNumericGroupId($gid);
```
jvillafanez: Isn't this a bit complex? Why do we want to use prefix?

PVince81:
- [ ] TODO: add unit tests
- [ ] TODO: remove LEN constant

jvillafanez:
Just a couple of extra things:

* From my point of view ( @DeepDiver1975 correct me if I'm wrong) there should be only one group manager and user manager, which are the ones provided by core. Creating and using another manager shouldn't be possible because both core and any app are interested in using the "official" manager.
* Custom backends, such as LDAP or this one could be registered in the core manager. This also includes the DB backend which is used by default in ownCloud.

PVince81:
@jvillafanez yes sorry, there is only one group manager.

What this PR implements is a group backend like LDAP does. (I mixed up both names...)

lib/CustomGroupsManager.php
jvillafanez: I don't think it's the backend responsability to resolve the collisions nor to prevent them. Any other backend could have a group with the same name "customgroup_12", and there shouldn't be any reason to avoid it.

I think it's core the one who should take care of how to resolve the collisions. Generally, just saying "I want this group" is enough to know what is the group you're refering to, but maybe sometimes you should say "I want this group which is located in this backend" so there is no doubt about the group you want.

PVince81: While I agree that it's not this backend's responsibility, we'll still get into trouble if all our groups are called "1", "2", "3", etc... And I don't intend to fix core to have its own collision detection / namespacing as part of this task, which is likely a bigger task on its own which might not be backward compatible.

lib/CustomGroupsManager.php
PVince81: For display names I'd say yes

PVince81:
- [ ] might need adjusting the whole PR depending on how we implement the new group backend interfaces as described in https://github.com/owncloud/core/issues/7918#issuecomment-263615014 for display name support

PVince81:
Anyway, my first goal with this PR was to experiment to find out whether sharing would work automatically with this approach. It did!

PVince81:
- [ ] now requires https://github.com/owncloud/core/pull/26750 for display name support and the new `IGroupBackend` interface

If you test this you'll see that the display name is correctly shown in the sharee list 😄

PVince81:
- [ ] TODO: every member can also be a group admin, need to change the data structure

PVince81:
Ok, I spent too much time already implementing the Sabre stuff but not enough writing actual tests for GroupsManager 😄 

Will do tomorrow so we can get this first PR merged soon.

PVince81:
I've added unit tests now for the group backend and group manager.
Note that for the tests to run properly you also need to checkout the matching core branch https://github.com/owncloud/core/pull/26750
(or ideally review that one first).

I've removed the DAV stuff from this PR into a separate PR: https://github.com/owncloud/customgroups/pull/7

Please review @DeepDiver1975 @jvillafanez

PVince81:
@DeepDiver1975 mind reviewing so I can continue with the DAV stuff ? Thanks

lib/CustomGroupsManager.php
```diff
@@ -0,0 +1,321 @@
...
+ */
+
+namespace OCA\CustomGroups;
+
+use OCP\IDBConnection;
+
+/**
+ * Manager for custom group
+ */
+class CustomGroupsManager {
```
jvillafanez: I'm questioning the purpose of this class... does it make the code too complex if we move this code to the backend?

Anyway, we should rename the class. The backend class will have a manager above (provided by core, where the backend will be added) and also below (this class) which is confusing. `CustomGroupsDBHandler` seems a better choice.

appinfo/database.xml
```diff
@@ -0,0 +1,80 @@
...
+				<length>64</length>
+			</field>
+			<field>
+				<name>is_admin</name>
+				<type>integer</type>
+				<default>1</default>
+				<notnull>false</notnull>
+			</field>
+			<index>
+				<name>cg_gid_uid_index</name>
```
jvillafanez: I'm not sure if we should add another index only for the user_id. I expect the queries for groups will use this index, but I don't think the queries for users will.

lib/Application.php
```diff
@@ -0,0 +1,38 @@
...
+class Application extends App {
+	public function __construct (array $urlParams = array()) {
+		parent::__construct('customgroups', $urlParams);
+	}
+
+	/**
+	 * Register the group manager
+	 */
+	public function registerGroupBackend() {
+		$backend = $this->getContainer()->query('\OCA\CustomGroups\CustomGroupsBackend');
```
jvillafanez: I don't know where this come from.

PVince81: What do you mean ? This method is called in appinfo/app.php

PVince81:
Restarted travis build now that master contains the code we need

lib/Application.php
jvillafanez: The `\OCA\CustomGroups\CustomGroupsBackend` needs to be register in the container, but I don't know where this registration is made.

PVince81: No, it doesn't. DI container power !

It works because the class name matches exactly so it's able to resolve it.
You only need explicit registration when the names different or when your constructor requires parameters that cannot be injected (ex: strings)

PVince81:
- [ ] rename CustomGroupsManager to something that sounds more like it's a DB accessor

lib/Application.php
jvillafanez: 😮 👏

PVince81:
- [ ] challenge: implement initial DB structure with doctrine migration

PVince81:
Looks like for the Mapper I can throw away almost all my code (and unit tests) and start over. I don't think the benefit of having the mapper pattern here is bringing any significant benefit at this point. (it might even introduce more hurdles for the API consumer)

I'll abandon this approach and simply rename the manager to `CustomGroupsDatabaseHandler`.

PVince81:
Raised https://github.com/owncloud/customgroups/issues/10 to use doctrine migrations for initial DB.

I'd like to get this PR here merged so I can move forward with the Sabre plugins / API. That is, unless there is a no-go in this PR.

PVince81:
Code coverage 404:
```
Uploading code coverage for repository "g/owncloud/customgroups" and revision "71b39e8571e3cba39d09c79143dabcb786640b7b"... Failed

{"error":{"code":404,"message":"Not Found"}}
```

@DeepDiver1975 do we need to create a project on that website maybe for that to work ?

PVince81:
Rebased to include the doctrine migration for the initial DB, deleted database.xml.

@jvillafanez I've renamed the class to `CustomGroupsDatabaseHandler`.
I hope the PR is acceptable now. Thanks

lib/CustomGroupsDatabaseHandler.php
```diff
@@ -0,0 +1,321 @@
...
+	}
+
+	/**
+	 * Deletes the group with the given id
+	 *
+	 * @param int $gid numeric group id
+	 * @return true if group was deleted, false otherwise
+	 */
+	public function deleteGroup($gid) {
+		// Delete the group-user relation
```
jvillafanez: I'd wrap this in a transaction. If we want to delete a group, we want both actions to be executed or none of them. Something in the middle it be would be an unexpected result.

lib/CustomGroupsDatabaseHandler.php
```diff
@@ -0,0 +1,321 @@
...
+		$qb = $this->dbConn->getQueryBuilder();
+		$cursor = $qb->select(['group_id', 'uri', 'display_name'])
+			->from('custom_group')
+			->where($qb->expr()->eq($field, $qb->createNamedParameter($numericGroupId)))
+			->execute();
+		$result = $cursor->fetch();
+		$cursor->closeCursor();
+
+		if (!$result) {
+			return null;
```
PVince81: because `$result` is `false`, not `null`...

PVince81:
Also, looks like sqlite not happy with my DB stuff. Will look into it. (Mysql worked fine)

PVince81:
Something is wrong with the unique constraint. Looks like SQLite set the primary key only on "group_id" instead of both "group_id" + "user_id" 😢

PVince81:
```
sqlite> pragma table_info(oc_custom_group_member);
cid|name|type|notnull|dflt_value|pk
0|group_id|INTEGER|1||1
1|user_id|VARCHAR(64)|1||0
2|is_admin|INTEGER|1|0|0
```

primary key is only set on `group_id` even though I asked it to set it on both:
```php
    $table->setPrimaryKey(['group_id', 'user_id']);
```

debugging....

PVince81:
grrr, this is the resulting SQL generated by Doctrine:
```
  $sql[2]                        = (string[156]) 'CREATE TABLE oc_custom_group_member (group_id INTEGER NOT NULL PRIMARY KEY AUTOINCREMENT, user_id VARCHAR(64) NOT NULL, is_admin INTEGER DEFAULT 0 NOT NULL)';
```

I didn't even ask for autoincrement... Looks like a Doctrine bug.

We do have a table in core with a similar schema and that one looks correct:
```
sqlite> pragma table_info(oc_group_user);
cid|name|type|notnull|dflt_value|pk
0|gid|VARCHAR(64)|1|''|1
1|uid|VARCHAR(64)|1|''|2
```

so this means SQLite supports multi-column primary keys but Doctrine doesn't seem to be able to create that.

Googling now, I hope there is maybe a missing flag or something in the API.

PVince81:
Facepalm, I forgot to remove "autoincrement" when copy-pasting the definitions.
Funnily enough, Mysql didn't complain.

PVince81:
@jvillafanez I fixed all your comments and additionally fixed the DB schema.

PVince81:
@DeepDiver1975 can you create the repo for clover coverage ? https://travis-ci.org/owncloud/customgroups/jobs/183218691#L831

jvillafanez:
👍

PVince81:
I'm going to remove the code coverage thing for now and let @DeepDiver1975 sort it out separately.

PVince81:
Raised here https://github.com/owncloud/customgroups/issues/12

Issue #3, opened 2016-12-03
Webdav API for groups management


PVince81:
- [x] custom groups
    - [x] list groups PROPFIND http://localhost/owncloud/remote.php/dav/customgroups/groups/
        - [x] property `oc:display-name`: group display name
        - [x] child node `MembershipNode`, name is the custom group URI
    - [x] rename custom group: PROPPATCH oc:display-name on http://localhost/owncloud/remote.php/dav/customgroups/groups/$customGroupUri
        - [x] only group admin can rename
    - [x] delete group: DELETE http://localhost/owncloud/remote.php/dav/customgroups/groups/$customGroupUri
        - [x] only group admin can delete
    - [x] create group: MKCOL http://localhost/owncloud/remote.php/dav/customgroups/groups/$customGroupUri
        - [x] throws MethodNotAllowed if already exists
        - [x] creator automatically becomes group admin and member
- [x] memberships:
    - [x] list members PROPFIND http://localhost/owncloud/remote.php/dav/customgroups/groups/$numericGroupId/
        - [x] child node "MembershipNode" name is user id
        - [x] property `oc:role` (int) whether this member is also a group admin
        - [x] only members can list members, others get 403 when listing collection contents
    - [x] add member: PUT http://localhost/owncloud/remote.php/dav/customgroups/groups/$numericGroupId/$userId
        - [x] only group admin can add members
    - [x] remove member: DELETE http://localhost/owncloud/remote.php/dav/customgroups/groups/$customGroupUri/$userId
        - [x] only group admin can remove members
        - [x] group admin cannot remove self if no other admin exists in the group
        - [x] a group member can remove themselves with the same API call to leave the group
    - [x] change admin role of a member: PROPPATCH http://localhost/owncloud/remote.php/dav/customgroups/groups/$customGroupUri/$userId of the `oc:role` property
- [x] list group memberships of a given user: PROPFIND http://localhost/owncloud/remote.php/dav/customgroups/users/$userId/$membership
    - [x] child node is `MembershipNode` where name is group id
    - [x] membership node has "oc:role" attribute
- [x] regular admins can do the same as group admins
- [ ] provider REPORT in addition to PROPFIND for pagination

PVince81:
- [ ] REQUIRES: support for apps to register sabre plugins: https://github.com/owncloud/core/issues/26195

DeepDiver1975:
I'd prefer to not work on the numeric ids on the dav api layer.
[...]

PR #7, opened 2016-12-06
DAV endpoint

PVince81:
- [x] REQUIRES: https://github.com/owncloud/customgroups/pull/1 to be merged

This PR is based on top of https://github.com/owncloud/customgroups/pull/1.

This adds DAV endpoints for https://github.com/owncloud/customgroups/issues/3

### DIFF
--- a/lib/Dav/CustomGroupsPlugin.php
+++ b/lib/Dav/CustomGroupsPlugin.php
@@ -1,0 +1,195 @@
+<?php
+/**
+ * @author Vincent Petry <pvince81@owncloud.com>
+ *
+ * @copyright Copyright (c) 2016, ownCloud GmbH
+ * @license AGPL-3.0
+ *
+ * This code is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License, version 3,
+ * as published by the Free Software Foundation.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License, version 3,
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>
+ *
+ */
+
+namespace OCA\CustomGroups\Dav;
+
+use OCA\CustomGroups\CustomGroupsDatabaseHandler;
+use OCP\IUserSession;
+use OCA\CustomGroups\Dav\RootCollection;
+use OCA\CustomGroups\Dav\MembershipNode;
+use OCA\CustomGroups\Dav\GroupsCollection;
+use Sabre\DAV\ServerPlugin;
+use OCA\CustomGroups\Dav\CustomGroupMemberNode;
+use Sabre\DAV\Exception\BadRequest;
+use Sabre\DAV\Xml\Element\Response;
+use Sabre\DAV\Xml\Response\MultiStatus;
+/**
+ * Sabre plugin to handle custom groups
+ */
+class CustomGroupsPlugin extends ServerPlugin {
+	const NS_OWNCLOUD = 'http://owncloud.org/ns';
+
+	const REPORT_NAME = '{http://owncloud.org/ns}filter-members';
+
+	/**
+	 * Custom groups handler
+	 *
+	 * @var CustomGroupsDatabaseHandler
+	 */
+	protected $groupsHandler;
+
+	/**
+	 * Sabre server
+	 *
+	 * @var \Sabre\DAV\Server $server
+	 */
+	private $server;
+
+	/**
+	 * User session
+	 *
+	 * @var \OCP\IUserSession
+	 */
+	protected $userSession;
+
+	/**
+	 * Custom groups plugin
+	 *
+	 * @param CustomGroupsDatabaseHandler $groupsHandler custom groups handler
+	 * @param IUserSession $userSession user session
+	 */
+	public function __construct(CustomGroupsDatabaseHandler $groupsHandler, IUserSession $userSession) {
+		$this->groupsHandler = $groupsHandler;
+		$this->userSession = $userSession;
+	}
+
+	/**
+	 * This initializes the plugin.
+	 *
+	 * This function is called by Sabre\DAV\Server, after
+	 * addPlugin is called.
+	 *
+	 * This method should set up the required event subscriptions.
+	 *
+	 * @param \Sabre\DAV\Server $server Sabre server
+	 */
+	public function initialize(\Sabre\DAV\Server $server) {
+		$this->server = $server;
+
+		$uri = $this->server->getRequestUri();
+		$uri = '/' . trim($uri) . '/';
+		if (strpos($uri, '/customgroups/') === false) {
+			return;
+		}
+
+		if ($this->userSession === null || $this->userSession->getUser() === null) {
+			return;
+		}
+
+		$this->server->xml->namespaceMap[self::NS_OWNCLOUD] = 'oc';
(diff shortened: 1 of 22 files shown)


PVince81:
Rebased, added lots of tests.

A few calls and conditions are missing still.

PVince81:
Already implemented the main part of the spec https://github.com/owncloud/customgroups/issues/3 for managing groups and memberships.

Next step: REPORT

PVince81:
I've added the REPORT as well:

Here is how to get one.
Put this into `report-customgroups.xml`:
```xml
<?xml version="1.0" encoding="utf-8" ?>
<oc:filter-members xmlns:a="DAV:" xmlns:oc="http://owncloud.org/ns" >
	<oc:filter-rules>
		<oc:user-id>user1</oc:user-id>
		<!--	<oc:is-admin>1</oc:is-admin> -->
	</oc:filter-rules>
</oc:filter-members>
```
Then `curl -u admin:admin -X REPORT -H "Content-Type: text/xml" --data-binary "@report-customgroups.xml" 'http://localhost/owncloud/remote.php/dav/customgroups/'`

Gives:
```xml
<?xml version="1.0"?>
<d:multistatus xmlns:d="DAV:" xmlns:s="http://sabredav.org/ns" xmlns:cal="urn:ietf:params:xml:ns:caldav" xmlns:cs="http://calendarserver.org/ns/" xmlns:card="urn:ietf:params:xml:ns:carddav" xmlns:oc="http://owncloud.org/ns">
 <d:response>
  <d:status>HTTP/1.1 200 OK</d:status>
  <d:href>/owncloud/remote.php/dav/customgroups/x/user1</d:href>
  <d:propstat>
   <d:prop>
    <oc:is-admin>0</oc:is-admin>
   </d:prop>
   <d:status>HTTP/1.1 200 OK</d:status>
  </d:propstat>
 </d:response>
 <d:response>
  <d:status>HTTP/1.1 200 OK</d:status>
  <d:href>/owncloud/remote.php/dav/customgroups/y/user1</d:href>
  <d:propstat>
   <d:prop>
    <oc:is-admin>0</oc:is-admin>
   </d:prop>
   <d:status>HTTP/1.1 200 OK</d:status>
  </d:propstat>
 </d:response>
</d:multistatus>
```

It basically returns the result for the `CustomGroupMemberNode`. I'm not too happy about this because it will not be useful for the UI. It misses the group display name and the group URI. Well, the group URI can be extracted from the result path.

I'd like to refine this separately once we start working on the UI.

PVince81:
- [ ] make it possible for regular admins to do the same things like group admins

The sad thing about adding this is that now I need to inject a group manager everywhere...

lib/Dav/CustomGroupsPlugin.php
```diff
@@ -0,0 +1,175 @@
...
+
+	/**
+	 * This initializes the plugin.
+	 *
+	 * This function is called by Sabre\DAV\Server, after
+	 * addPlugin is called.
+	 *
+	 * This method should set up the required event subscriptions.
+	 *
+	 * @param Server $server
```
DeepDiver1975: Server is unknown

lib/Dav/CustomGroupsPlugin.php
```diff
@@ -0,0 +1,175 @@
...
+	}
+
+	/**
+	 * REPORT operations to look for comments
+	 *
+	 * @param string $reportName
+	 * @param [] $report
+	 * @param string $uri
+	 * @return bool
+	 * @throws NotFound
```
DeepDiver1975: NotFound is unknown

lib/Dav/CustomGroupsPlugin.php
```diff
@@ -0,0 +1,175 @@
...
+	 * @throws ReportNotSupported
```
DeepDiver1975: ReportNotSupported is unknown

lib/Dav/CustomGroupsPlugin.php
```diff
@@ -0,0 +1,175 @@
...
+	 */
+	public function onReport($reportName, $report, $uri) {
+		$node = $this->server->tree->getNodeForPath($uri);
+		if(!$node instanceof RootCollection || $reportName !== self::REPORT_NAME) {
+			throw new ReportNotSupported();
```
DeepDiver1975: ReportNotSupported is unknown

lib/Dav/RootCollection.php
```diff
@@ -0,0 +1,169 @@
+<?php
+/**
+ * @author Vincent Petry <pvince81@owncloud.com>
+ *
+ * @copyright Copyright (c) 2016, ownCloud GmbH.
```
DeepDiver1975: remove the .

lib/Dav/RootCollection.php
```diff
@@ -0,0 +1,169 @@
...
+	 * @var IUserManager
+	 */
+	private $userManager;
+
+	/**
+	 * @var IUserSession
+	 */
+	private $userSession;
+
+	/**
```
DeepDiver1975: PHPDoc missing

lib/Dav/CustomGroupMemberNode.php
```diff
@@ -0,0 +1,250 @@
...
+				self::PROPERTY_IS_ADMIN => $this->memberInfo['is_admin'] ? 1 : 0,
+			];
+		}
+		return [];
+	}
+
+	/**
+	 * Update the admin flag
+	 *
+	 * @param int $isAdminPropValue 1 for true, 0 for false
```
DeepDiver1975: return tag is missing

DeepDiver1975:
> It basically returns the result for the CustomGroupMemberNode. I'm not too happy about this because it will not be useful for the UI. It misses the group display name and the group URI. Well, the group URI can be extracted from the result path.

please have a look at the carddav query request - there is a node where you select the elements which you expect to be returned ...

https://tools.ietf.org/html/rfc6352#section-8.6.3

```xml
     <D:prop>
       <D:getetag/>
       <CG:display/>
     </D:prop>

PVince81:
Squashed and removed the composer files I checked in by mistake from the other branch.

PVince81:
ownCloud admins can now also manage groups even when they aren't members.

Instead of passing around yet another object `IGroupManaber`, I created a helper class instead `MembershipHelper` which takes care of checking whether the current user can or cannot manage stuff based on their role and memberships.

PVince81:
@DeepDiver1975 please re-review.

My next step is to adjust the REPORT to implement pagination, but this could also be done separately if this PR is already mergable.

PVince81:
I'm feeling a bit lazy with the REPORT + pagination especially if I need to do it for all the different endpoints. So I came up with this generic paginated report idea, please check it out: https://github.com/owncloud/core/issues/26840

PVince81:
REPORT will be done in a separate PR.

@DeepDiver1975 please review

DeepDiver1975:
- [ ] listing users does not work /remote.php/dav/customgroups/users/

PVince81:
> listing users does not work /remote.php/dav/customgroups/users/

This is by design, same like the "users" list principal.

You need to know the user name you want to query.

DeepDiver1975:
> This is by design, same like the "users" list principal.

Okay - in debug mode we allow listing of users in other collections as well

PVince81:
> Okay - in debug mode we allow listing of users in other collections as well

Yet another code path to test / unit test and maintain. I'd rather not.

Unless we make it more useful by enabling user listing for any admins
And even then, the admin should be able to get a listing from the other endpoint.

Let me know what you think.

PVince81:
Adjusted exception formatting.

Added missing case: ownCloud admin can now enter the "users/$userId" subcollection even if not a member.

@DeepDiver1975 I'd like to not add user listing here, it would require injecting the user manager, etc. I don't think it's useful enough, especially if only used in debug mode.

Issue #10, opened 2016-12-09
Use Doctrine Migration for initial DB structure


PVince81:
With https://github.com/owncloud/core/pull/14851

But met some issues so far: https://github.com/owncloud/core/pull/14851#issuecomment-265987281

Goal is mostly to test migrations.

PR #11, opened 2016-12-09
Initialize DB with migration

PVince81:
Fixes https://github.com/owncloud/customgroups/issues/10

### DIFF
--- a/appinfo/Migrations/Version20161209151129.php
+++ b/appinfo/Migrations/Version20161209151129.php
@@ -1,0 +1,71 @@
+<?php
+
+namespace OCA\CustomGroups\Migrations;
+
+use Doctrine\DBAL\Migrations\AbstractMigration;
+use Doctrine\DBAL\Schema\Schema;
+
+/**
+ * Auto-generated Migration: Please modify to your needs!
+ */
+class Version20161209151129 extends AbstractMigration {
+	/**
+	 * @param Schema $schema
+	 */
+	public function up(Schema $schema) {
+		$this->createGroupsTable($schema);
+		$this->createMembersTable($schema);
+	}
+
+	private function createGroupsTable(Schema $schema) {
+		$prefix = $this->connection->getPrefix();
+		$table = $schema->createTable("${prefix}custom_group");
+		$table->addColumn('group_id', 'integer', [
+			'autoincrement' => true,
+			'unsigned' => true,
+			'length' => 4,
+		]);
+		$table->addColumn('uri', 'text', [
+			'length' => 255,
+			'notnull' => true,
+		]);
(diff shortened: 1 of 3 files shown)


PVince81:
I'm stuck because am unable to create a unique index.
In MDB2 schema we had this:
```xml
<index>
    <name>cg_uri_index</name>
    <unique>true</unique>
    <field>
        <name>uri</name>
        <sorting>ascending</sorting>
    </field>
</index>
```

Now with Doctrine I define it as:
```php
    $table->addUniqueIndex(['uri'], 'cg_uri_index');
```

But MySQL is not happy:
```
± % sudo -uwwwrun ./occ app:enable customgroups
                              
  [Doctrine\DBAL\Exception\DriverException]                                                                                                                                                 
  An exception occurred while executing 'CREATE TABLE oc_custom_group (group_id INT UNSIGNED AUTO_INCREMENT NOT NULL, uri TINYTEXT NOT NULL, display_name TINYTEXT NOT NULL, UNIQUE INDEX   
  cg_uri_index (uri), PRIMARY KEY(group_id)) DEFAULT CHARACTER SET utf8 COLLATE utf8_unicode_ci ENGINE = InnoDB':                                                                           
  SQLSTATE[42000]: Syntax error or access violation: 1170 BLOB/TEXT column 'uri' used in key specification without a key length                                                             
                                                                                                                                                                                            

                                                                                                                                 
  [Doctrine\DBAL\Driver\PDOException]                                                                                            
  SQLSTATE[42000]: Syntax error or access violation: 1170 BLOB/TEXT column 'uri' used in key specification without a key length  
                                                                                                                                 

                                                                                                                                 
  [PDOException]                                                                                                                 
  SQLSTATE[42000]: Syntax error or access violation: 1170 BLOB/TEXT column 'uri' used in key specification without a key length  
```

This says that MySQL requires an explicit key length, basically how many of first characters of the field must be indexed: http://stackoverflow.com/a/1827118

However Doctrine doesn't allow you to pass a length for indices: https://github.com/doctrine/doctrine2/issues/3553 (won't fix as too DB specific)

I was told on IRC to try with varchar, let's find out how to do that...

PVince81:
Note that in the MDB2 schema the "uri" column was specified as having the type "text" which is somehow mapped to "TINYTEXT"

PVince81:
I changed the column type from "text" to "string" and now the index can be created properly.

It is now possible to enable this app and get the initial DB !

@DeepDiver1975 @VicDeo @jvillafanez please review

PVince81:
btw, requires https://github.com/owncloud/core/pull/14851 checked out on master for it to work

jvillafanez:
👍

PR #13, opened 2016-12-14
Preliminary frontend work

PVince81:
Preliminary work for the frontend.

This extends the Makefile significantly and adds the following:
- composer deps (currently only phpunit)
- nodejs deps (karma + bower + handlebars)
- bower deps
- karma test runner
- handlebars compilation


PVince81:
- [ ] REQUIRES: enhance oc-backbone-webdav adapter to support MKCOL for creating collections (default is POST or PUT)

PVince81:
- [ ] update handlebars version in core to avoid conflict, it will be needed anyway for future handlebars precompiling (https://github.com/owncloud/core/pull/26021/files#diff-0a08a7565aba4405282251491979bb6bR17 https://github.com/owncloud/core/pull/26021/files#diff-c5776441e263bb28be0b4b6e51fd527fR23)

PVince81:
Rebased.

Will require adjustments to be able to work with the updated endpoint structure.

PVince81:
This "uri" approach is a nightmare.

Backbone (and potentially other frameworks) assume that if the id is known, then the object already exists. I have used the "uri" as id so it always uses the "update" internal method instead of "create" and tries to PROPPATCH instead of MKCOL.

I think we'll still need to use the actual internal id field as the backbone id, and the uri is only used for the path. I'll try that approach then.

PVince81:
- [ ] REQUIRES: https://github.com/owncloud/core/pull/27119 to be able to create with MKCOL
- [ ] REQUIRES: https://github.com/owncloud/core/pull/27118 to be able to rename

PVince81:
- [ ] BUG: the Webdav endpoint must return the internal ID as header after MKCOL, for Backbone to be able to set an internal id

PVince81:
So I finally have a solution, but it required me to write custom model and collections specifically for Webdav. This would make it reusable in the future. See https://github.com/owncloud/core/pull/27119.

The frontend now works, you can now:
- create/rename/delete groups
- leave a group
- rename a group
- add/delete members
- switch role for a member

PVince81:
Latest state:
Custom Groups

PVince81:
This is now ready for review.

Note that this core PR is still required to test this: https://github.com/owncloud/core/pull/27119 (make sure to run `make` in core again to get the updated lib)

Everything is squashed into two commits:
The first commit is about improving the base stuff with a better Makefile (try `make help`). I hope to make this reusable for other apps.
The second commit adds all the JS code + JS unit tests to make the UI work as seen in https://github.com/owncloud/customgroups/pull/13#issuecomment-284714384

Please review @felixheidecke @jvillafanez @DeepDiver1975 @VicDeo @IljaN

PVince81:
... and if you're wondering why I used `window.prompt` and `window.confirm`:
this here is the crude version of the UI. It is functional, but doesn't provide the best UX yet.
UX will be improved in subsequent PRs.

At least we could say this is a POC that could be used for demoing the functionality.

PVince81:
ah of course, unit tests cannot pass until https://github.com/owncloud/core/pull/27119 from core is merged

lib/Dav/CustomGroupsPlugin.php
```diff
@@ -91,7 +91,8 @@ public function initialize(\Sabre\DAV\Server $server) {
 		$this->server->xml->namespaceMap[self::NS_OWNCLOUD] = 'oc';
 		$ns = '{' . self::NS_OWNCLOUD . '}';
 		$this->server->resourceTypeMapping[MembershipNode::class] = $ns . 'customgroups-membership';
-		$this->server->resourceTypeMapping[GroupsCollection::class] = $ns . 'customgroups-group';
+		$this->server->resourceTypeMapping[GroupsCollection::class] = $ns . 'customgroups-groups';
```
jvillafanez: "customgroups-groups" or "customgroups-group"? Is this intended?

lib/Dav/MembershipNode.php
```diff
@@ -36,7 +36,6 @@ class MembershipNode implements \Sabre\DAV\INode, \Sabre\DAV\IProperties {
 
 	const PROPERTY_ROLE = '{http://owncloud.org/ns}role';
 	const PROPERTY_USER_ID = '{http://owncloud.org/ns}user-id';
-	const PROPERTY_GROUP_URI = '{http://owncloud.org/ns}group-uri';
```
jvillafanez: Any reason to remove this? Mainly to confirm this isn't accidental

jvillafanez:
I've also seen that the windows prompts aren't translatable... I think that we should make them translatable in order to merge this, so we can change that part at any time. Otherwise we might need to rush to change the implementation if this is merged.

jvillafanez:
I haven't reviewed the tests yet.... 😓

SergioBertolinSG:
When adding new users to custom group there is no autocomplete in the textfield. This will come after this PR or it is a bug here?

SergioBertolinSG:
When editing a member role, just a click on the pencil change member's role. Feels a bit unexpected, I think it should be a dropdown with the options admin or member.

lib/Controller/PageController.php
```diff
@@ -0,0 +1,40 @@
...
+
+class PageController extends Controller {
+
+	/**
+	 * @NoCSRFRequired
+	 * @NoAdminRequired
+	 */
+	public function index() {
+		// TODO: cache or add to info.xml ?
+		$modules = json_decode(file_get_contents(__DIR__ . '/../../js/modules.json'));
```
PVince81: The second usage is in the unit tests, that's why I wanted a separate file... I'm still longing for a proper portable way to load JS files for OC apps.

js/templates/membersList.handlebars
```diff
@@ -0,0 +1,13 @@
+<div class="header">
+</div>
+<table class="grid">
+	<thead>
+		<th></th>
```
PVince81: This header is for the avatar column

js/GroupsView.js
```diff
@@ -0,0 +1,282 @@
...
+				model.get('displayName'),
+				32
+			);
+		},
+
+		_onRemoveGroup: function(model, collection, options) {
+			var $groupEl = this.$el.find('.group-list .group:nth-child(' + (options.index + 1) + ')');
+			if ($groupEl.hasClass('active')) {
+				// deselect
+				this.trigger('select', null);
```
PVince81: No, the only listener for select events is in the "App.js". Selecting nothing would make it close the sidebar.

PVince81:
> windows prompts

You mean the native window.prompt ? We need to replace these anyway... I put these here to save some dev time. They definitely need a better replacement. (separately)

PVince81:
@SergioBertolinSG I've raised them separately:
- autocomplete to be added later: https://github.com/owncloud/customgroups/issues/32
- permission change should be a dropdown instead of pencil: https://github.com/owncloud/customgroups/issues/33

js/App.js
```diff
@@ -0,0 +1,89 @@
...
+		template: function(data) {
+			return OCA.CustomGroups.Templates.app(data);
+		},
+
+		render: function() {
+			this.$el.html(this.template({
+				customGroupsTitle: t('customgroups', 'Custom Groups')
+			}));
+
+			this.$groupsContainer = this.$('.groups-container').removeClass('icon-loading');
```
PVince81: `this.$('.groups-container')` is a Backbone View shortcut for `this.$el.find('.groups-container')`

As for `this.$groupsContainer` the `$` sign is a convention that already tells that this attribute is a jQuery object, just like `this.$el`.

Hope that makes sense.

js/App.js
```diff
@@ -0,0 +1,89 @@
...
+			this.$membersContainer = this.$('.members-container').removeClass('icon-loading');
+
+			this.listView.render();
+			this.$groupsContainer.append(this.listView.$el);
+		},
+
+		_onSelectGroup: function(group) {
+			this.$membersContainer.empty();
+			var state = {
```
PVince81: forgot to kill it...

js/GroupsView.js
```diff
@@ -0,0 +1,282 @@
...
+			this.collection.on('sync', this._onEndRequest, this);
+			this.collection.on('add', this._onAddModel, this);
+			this.collection.on('change', this._onChangeModel, this);
+			this.collection.on('remove', this._onRemoveGroup, this);
+
+			_.bindAll(
+				this,
+				'_onSubmitCreationForm',
+				'_onRenameGroup',
+				'_onDeleteGroup'
```
PVince81: actually I'll remove them all because backbone auto-binds when specified in the "events" block

js/GroupsView.js
```diff
@@ -0,0 +1,282 @@
...
+				return false;
+			}
+
+			// TODO: use inline rename form
+			var newName = window.prompt('Enter new name', model.get('displayName'));
+			if (newName) {
+				// TODO: lock row during save
+				model.save({
+					displayName: newName
+				});
```
PVince81: no, this is a jquery event handler and the return value indicates whether bubbling is needed (or something like this).

Actually we don't need to return because we already call `ev.preventDefault();` at the top.

I'll still add add `return false;`

js/GroupsView.js
```diff
@@ -0,0 +1,282 @@
...
+			var model = this._findGroupModelForEvent(ev); 
+			if (!model) {
+				return false;
+			}
+
+			// TODO: use undo approach
+			if (window.confirm('Confirm deletion of group ' + model.get('displayName') + ' ?')) {
+				model.destroy({
+					wait: true
+				});
```
PVince81: same as above, jquery event handler

js/GroupsView.js
```diff
@@ -0,0 +1,282 @@
...
+					self.$('[name=groupName]').val('');
+					self.select(model);
+				},
+				error: function(model, response) {
+					if (response.status === 405) {
+						if (_.isUndefined(index)) {
+							// attempt again with index
+							index = 1;
+						}
+						self._createGroup(groupName, index + 1);
```
PVince81: To be honest I dislike having to do this on the client side... But for the sake of Webdav-ness... Oh well.

Let's stop at 100 then

js/GroupsView.js
```diff
@@ -0,0 +1,282 @@
...
+			var $el = $(this.itemTemplate(this._formatItem(model)));
+			var index = this.collection.indexOf(model);
+			this.$container.children().eq(index).replaceWith($el);
+			this._postProcessRow($el, model);
+		},
+
+
+		_postProcessRow: function($el, model) {
+			$el.find('[title]').tooltip();
+			$el.find('.avatar').imageplaceholder(
```
PVince81: The `avatar` method would attempt to fetch a group avatar from the server side. We don't have them so it will always result in 404. So I took the shortcut and use `imageplaceholder` directly.

I'll add a comment to clarify.

js/GroupsView.js
```diff
@@ -0,0 +1,282 @@
...
+				// combine uri with display name for seed
+				model.id + ':' + model.get('displayName'),
+				model.get('displayName'),
+				32
+			);
+		},
+
+		_onRemoveGroup: function(model, collection, options) {
+			var $groupEl = this.$el.find('.group-list .group:nth-child(' + (options.index + 1) + ')');
```
PVince81: I'll add a fallback to filtering on the "data-id" attribute. The reason I picked index is that it is likely more efficient to resolve.

js/GroupsView.js
```diff
@@ -0,0 +1,282 @@
...
+				return groupName + index;
+			}
+			return groupName;
+		},
+
+		_formatItem: function(group) {
+			return {
+				id: group.id,
+				displayName: group.get('displayName'),
+				renameLabel: t('customgroups', 'Rename'),
```
PVince81: It is yes... I'm still hoping to be able to put translated texts directly into templates, requires https://github.com/owncloud/core/issues/15106

js/MembersView.js
```diff
@@ -0,0 +1,268 @@
...
+			this.collection.off('sync', this._onEndRequest, this);
+			this.collection.off('add', this._onAddModel, this);
+			this.collection.off('change', this._onChangeModel, this);
+			this.collection.off('remove', this._onRemoveMember, this);
+			this.model.off('change:displayName', this._renderHeader, this);
+
+			this.$el.remove();
+		},
+
+		_toggleLoading: function(state) {
```
PVince81: In the back of my mind I was thinking of maybe having a base class `ListView` with all that list logic in core and then extend it in apps... reusable components... we'll see in the future

js/MembersView.js
```diff
@@ -0,0 +1,268 @@
...
+			this.$el.find('.empty').toggleClass('hidden', !!this.collection.length);
+		},
+
+		_onClickLeaveGroup: function() {
+			var currentUserMembership = this.collection.get(OC.getCurrentUser().uid);
+			if (window.confirm('Confirm leaving of group ' + this.model.get('displayName') + ' ?')) {
+				currentUserMembership.destroy({
+					wait: true,
+					error: function() {
+						OC.Notification.showTemporary(t('customgroups', 'Cannot leave group without an administrator'));
```
PVince81: indeed... will adjust

js/MembersView.js
```diff
@@ -0,0 +1,268 @@
...
+			this._postProcessRow($el, model);
+		},
+
+		_postProcessRow: function($el, model) {
+			$el.find('[title]').tooltip();
+			$el.find('.avatar').avatar(model.id, 32);
+		},
+
+		_onRemoveMember: function(model, collection, options) {
+			this.$el.find('.grid .group-member:nth-child(' + (options.index + 1) + ')').remove();
```
PVince81: will add fallback to `.filterAttr('data-id', model.id)`

js/templates/list.handlebars
```diff
@@ -0,0 +1,27 @@
+<form name="customGroupsCreationForm">
+	<div>
+		<input
```
PVince81: ok. now it has 134 columns 😉

lib/Controller/PageController.php
PVince81: let's discuss in https://github.com/owncloud/customgroups/issues/34

lib/Dav/CustomGroupsPlugin.php
PVince81: This is intended, but not specifically for the frontend work... just a "pass-by" fix.

"customgroups-groups" plural is for GroupsCollection as the collection is a list of groups, so "groups".
The next item GroupMemberShipCollection is actually a single group, so it has the type "customgroups-group" singular.

Still, thanks for spotting

lib/Dav/MembershipNode.php
PVince81: The property isn't used at all in this node.
It was a copy-paste from another node I believe so I removed it.

PVince81:
Addressed all review comments and squashed the fixes into the last commit.

Note that I also renamed the unit tests from CustomGroupsCollection to GroupsCollection now.
Originally I had "CustomGroups" as prefix everywhere but it became boring so I renamed the JS files but forgot to rename the test files. This is done here now.

@jvillafanez another check ?

js/App.js
jvillafanez: 👍

js/GroupsView.js
jvillafanez: 👍

js/GroupsView.js
jvillafanez: 👍

js/MembersView.js
jvillafanez: fair enough 👍

js/templates/membersList.handlebars
jvillafanez: maybe an HTML comment here to make sure these header are in use? Just looking exclusively at the file it seems they can be deleted.

lib/Controller/PageController.php
jvillafanez: 👍

js/templates/membersList.handlebars
PVince81: nah... will only do if I need to retouch this PR or next time I come around this place

PVince81: an experienced HTML dev would know that one does not simply remove an empty "th" or "td" because that would mess up the layout 😉

lib/Dav/CustomGroupsPlugin.php
jvillafanez: 👍

lib/Dav/MembershipNode.php
jvillafanez: 👍

jvillafanez:
> You mean the native window.prompt ? We need to replace these anyway... I put these here to save some dev time. They definitely need a better replacement. (separately)

Yes. My point is that I don't know how much time we have to change them. If we make their text translatable, we could release the app. It's ugly, nobody likes it, but it's usable. We can improve the app later. (I'm refering to the windows prompts, not the app itself 😄 )
If we merge it as it is, we'll have to change it before releasing the app because those strings aren't translatable, and we shouldn't release the app under those conditions.

js/GroupsView.js
jvillafanez: Maybe 100 is too many, but ok 👍

PVince81:
We definitely shouldn't release the final app with the ugly window prompts.
Raised https://github.com/owncloud/customgroups/issues/36

PR #17, opened 2017-02-06
Added infrastructure for custom groups testing

SergioBertolinSG:


PVince81:
Looks fine 👍 

I'll add a Makefile later

SergioBertolinSG:
Problem with changing autoloading:      

Adding two elements with hyphens it fails:

- '': %paths.base%/../../../../../build/integration/features/bootstrap
- '': %paths.base%/../features/bootstrap

Result:
```php
[Symfony\Component\Config\Definition\Exception\InvalidTypeException]          
  Invalid type for path "testwork.autoload.0". Expected scalar, but got array.
```

Without hyphens it loads only the content from the first path, not the second.

We need to figure out how to autoload contexts from both. 

Perhaps using profiles?
Perhaps using composer's PSR-4 autoload mechanism?

SergioBertolinSG:
This https://github.com/owncloud/core/pull/27104 has to be merged before travis can pass.

.travis.yml
```diff
@@ -28,16 +28,19 @@ before_install:
...
+  # Run integration tests
+  - sh -c "if [ '$INTEGRATION' = '1' ]; then cd tests/integration; ./run.sh; fi"
+
   # Test the app
-  - sh -c "if [ '$JSTESTS' != '1' -a '$CODECHECK' = '1' ]; then make test-codecheck; fi"
-  - sh -c "if [ '$JSTESTS' != '1' -a '$CODECHECK' = '1' ]; then make test-codecheck; fi"
-  - sh -c "if [ '$JSTESTS' != '1' -a '$CODECHECK' = '2' ]; then make test-codecheck-deprecations; fi"
+  - sh -c "if [ '$JSTESTS' != '1' -a '$CODECHECK' = '1' -a '$INTEGRATION' != '1' ]; then make test-codecheck; fi"
+  - sh -c "if [ '$JSTESTS' != '1' -a '$CODECHECK' = '1' -a '$INTEGRATION' != '1' ]; then make test-codecheck; fi"
+  - sh -c "if [ '$JSTESTS' != '1' -a '$CODECHECK' = '2' -a '$INTEGRATION' != '1' ]; then make test-codecheck-deprecations; fi"
```
PVince81: It starts to feel like we should have only a single variable "TEST_WHATTOTEST=integration" or other values.

Anyway, something for another time.

PVince81:
Great start 👍

PR #19, opened 2017-02-14
Integration tests adding more basic cases

SergioBertolinSG:
Currently the renaming test is failing:

```
curl -X PROPPATCH -H "Content-Type: text/xml" --data-binary "@proppatch-customgroups.txt" http://admin:admin@HOSTNAME/remote.php/dav/customgroups/groups/group0
```

```
<?xml version="1.0" encoding="utf-8"?>
<d:error xmlns:d="DAV:" xmlns:s="http://sabredav.org/ns">
  <s:exception>UnexpectedValueException</s:exception>
  <s:message>A callback sent to handle() did not return an array or a bool</s:message>
</d:error>
```

This requires a rebase after fix arrives to master.

### DIFF
--- a/tests/integration/custom_groups_features/CustomGroups.feature
+++ b/tests/integration/custom_groups_features/CustomGroups.feature
@@ -4,9 +4,19 @@ Background:
 		Given using api version "1"
 		And using new dav path
 
-Scenario: Create a group
+Scenario: Create a custom group
 		Given As an "admin"
 		And user "user0" exists
 		When user "user0" created a custom group called "group0"
-		And the HTTP status code should be "201"
+		Then the HTTP status code should be "201"
 		And custom group "group0" exists
+
+
+Scenario: Rename a custom group
+		Given As an "admin"
+		And user "user0" exists
+		And user "user0" created a custom group called "group0"
+		And custom group "group0" exists
+		When user "user0" renamed custom group "group0" as "renamed-group0"
+		Then the HTTP status code should be "201"
+		And custom group "renamed-group0" exists
--- a/tests/integration/custom_groups_features/CustomGroups.feature
+++ b/tests/integration/custom_groups_features/CustomGroups.feature
@@ -134,3 +134,15 @@ Scenario: Only an admin of a custom group can remove members
 					| user0 |
 					| member1 |
 					| member2 |
+
+Scenario: Group admin cannot remove self if no other admin exists in the group
+		Given As an "admin"
+		And user "user0" exists
+		And user "member1" exists
+		And user "user0" created a custom group called "group0"
+		And user "user0" maked user "member1" member of custom group "group0"
+		When user "user0" removed membership of user "user0" from custom group "group0"
+		Then the HTTP status code should be "403"
+		And members of "group0" requested by user "user0" are
+					| user0 |
+					| member1 |
--- a/tests/integration/custom_groups_features/CustomGroups.feature
+++ b/tests/integration/custom_groups_features/CustomGroups.feature
@@ -166,3 +166,18 @@ Scenario: A member of a custom group can leave the custom group himself
 		Then the HTTP status code should be "204"
 		And members of "group0" requested by user "user0" are
 					| user0 |
+
+Scenario: A user can list his groups
+		Given As an "admin"
+		And user "user0" exists
+		And user "member1" exists
+		And user "user0" created a custom group called "group0"
+		And user "user0" created a custom group called "group1"
+		And user "user0" created a custom group called "group2"
+		When user "user0" made user "member1" member of custom group "group0"
+		When user "user0" made user "member1" member of custom group "group1"
+		When user "user0" made user "member1" member of custom group "group2"
+		Then custom groups of "member1" requested by user "member1" are
+					| group0 |
+					| group1 |
+					| group2 |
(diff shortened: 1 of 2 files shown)


tests/integration/custom_groups_features/CustomGroups.feature
```diff
@@ -134,3 +134,15 @@ Scenario: Only an admin of a custom group can remove members
 					| user0 |
 					| member1 |
 					| member2 |
+
+Scenario: Group admin cannot remove self if no other admin exists in the group
+		Given As an "admin"
+		And user "user0" exists
+		And user "member1" exists
+		And user "user0" created a custom group called "group0"
+		And user "user0" maked user "member1" member of custom group "group0"
```
PVince81: `maked` => `made`

PVince81: or `makes`

tests/integration/custom_groups_features/CustomGroups.feature
```diff
@@ -4,9 +4,245 @@ Background:
...
+
+Scenario: Rename a custom group
+		Given As an "admin"
+		And user "user0" exists
+		And user "user0" created a custom group called "group0"
+		And custom group "group0" exists
+		When user "user0" renamed custom group "group0" as "renamed-group0"
+		Then custom group "group0" exists with display name "renamed-group0"
+
+Scenario: Only a custom group admin can rename its custom group
```
PVince81: is there a test that an error code is returned if a non-admin attempts this ?

PVince81: ah, never mind. was looking for an error code but the test itself checks that there was no change

tests/integration/custom_groups_features/CustomGroups.feature
```diff
@@ -4,9 +4,245 @@ Background:
...
+		And user "member2" exists
+		And user "user0" created a custom group called "group0"
+		And user "user0" made user "member1" member of custom group "group0"
+		When user "member1" made user "member2" member of custom group "group0"
+		Then the HTTP status code should be "403"
+		And members of "group0" requested by user "user0" are
+					| user0 |
+					| member1 |
+
+Scenario: Only a custom group member can list members
```
PVince81: To avoid confusion, please rename the "Only ..." cases and call them something like "A non-member cannot list members". Because the current makes it think that the test will also cover listing members for custom group member but actually the test is there to test that someone **cannot** do something. So better have the word "cannot" in the scenario description.

tests/integration/custom_groups_features/CustomGroups.feature
```diff
@@ -4,9 +4,245 @@ Background:
...
+		And user "user0" created a custom group called "group2"
+		When user "user0" made user "member1" member of custom group "group0"
+		And user "user0" made user "member1" member of custom group "group1"
+		And user "user0" made user "member1" member of custom group "group2"
+		Then custom groups of "member1" requested by user "member1" are
+					| group0 |
+					| group1 |
+					| group2 |
+
+Scenario: Change role of a member of a group
```
PVince81: one missing scenario: "an admin cannot remove their own admin permissions if there is no other admin in the group"

PVince81: meaning "change their role to 'member'", not leave the group as per your previous test above

tests/integration/features/bootstrap/CustomGroupsContext.php
```diff
@@ -77,13 +96,250 @@ public function customGroupExists($customGroup){
...
+	 * @Then custom group :customGroup exists with display name :displayName
+	 * @param string $customGroup
+	 * @param string $displayName
+	 */
+	public function customGroupExistsWithDisplayName($customGroup, $displayName){
+		$customGroupsList = $this->getCustomGroups("admin");
+		$exists = false;
+		foreach($customGroupsList as $customGroupPath => $customGroupName) {
+			if ((!empty($customGroupName)) && (array_values($customGroupName)[0] == $displayName)){
+				$exists = true;
```
PVince81: you can call `break;` here

tests/integration/custom_groups_features/CustomGroups.feature
SergioBertolinSG: It checks that the group name is still the same, not the error code answered.

PVince81:
Looks good overall. Please have a look at my minor comments.

tests/integration/custom_groups_features/CustomGroups.feature
SergioBertolinSG: Added scenario.

tests/integration/features/bootstrap/CustomGroupsContext.php
SergioBertolinSG: Done.

tests/integration/features/bootstrap/CustomGroupsContext.php
```diff
@@ -106,6 +106,7 @@ public function customGroupExistsWithDisplayName($customGroup, $displayName){
 		foreach($customGroupsList as $customGroupPath => $customGroupName) {
 			if ((!empty($customGroupName)) && (array_values($customGroupName)[0] == $displayName)){
 				$exists = true;
+				break;
```
PVince81: there are other similar code pieces where you can use `break`. Anyone, very minor, will likely save only a microsecond when running the tests

PVince81:
:+1:

Should we squash on merge or is there any value in keeping all your commits separate ?

SergioBertolinSG:
I think they are more expressive separated. But as you wish.

PR #21, opened 2017-02-16
Backend fixes

PVince81:
Some fixes for the way how the backend works:

- role is now returned as string through the DAV property: "member" or "admin"
- setting an invalid role will return 400 for that specific property
- it is now possible to access and manage groups through the "users" endpoint:
   - PROPFIND "dav/customgroups/users/user1/" to list all groups in which "user1" is member or admin in
   - MKCOL "dav/customgroups/users/user1/newgroup"
   - PROPFIND "dav/customgroups/users/user1/newgroup" to list members, etc
- the "dav/customgroups/groups/" now only to be used by the superadmin

The latter endpoint changes match better with the frontend requirements: any user has their own list of groups in which they are members. They cannot see all groups in that list, only the ones they are admin or member in.

@SergioBertolinSG @DeepDiver1975 please review.


SergioBertolinSG:
When I try to get all members of a group I am getting an error:

txt file is:
```xml
<?xml version="1.0"?>
<a:propfind xmlns:a="DAV:" xmlns:oc="http://owncloud.org/ns">
<a:prop><oc:role/></a:prop>
</a:propfind>
```

Request
```
curl -X PROPFIND  -H "Content-Type: text/xml" --data-binary "@propfind_members_custom_groups.txt" http://user0:user0@HOST/remote.php/dav/customgroups/groups/group0
```
     
Answer
```xml
<?xml version="1.0" encoding="utf-8"?>
<d:error xmlns:d="DAV:" xmlns:s="http://sabredav.org/ns">
  <s:exception>InvalidArgumentException</s:exception>
  <s:message>Invalid backend role ""</s:message>
</d:error>
```

PVince81:
Fixed. Turns out I missed a piece when extracting the changes from the frontend branch.

SergioBertolinSG:
> the "dav/customgroups/groups/" now only to be used by the superadmin

Not sure what is the definition of superadmin here. Currently admins of custom groups can get what custom groups are in the system.

PVince81:
Superadmin is basically the ownCloud admin. Any user who is part of the "admin" system group. These can do everything. This way it's convenient for such admin to clean up old groups of departed people.

SergioBertolinSG:
So I guess it is not working as expected since every regular user but admin of a custom group can see all custom groups.

PVince81:
@SergioBertolinSG it depends on the endpoint that you use.

The thing about custom groups is that it's ok for regular users to see about all groups, because they will be able to share with said groups even if they're not members. This is why the "groups/" endpoint allows seeing all, but not browseing inside.

However the UI (from the frontend branch) is built using the second endpoint "users/$userId/" which only displays the groups in which the user is member of.

Now one can wonder why we even the "groups/" endpoint in the first place. Not sure right now... apart from the superadmin it doesn't seem useful.

PR #26, opened 2017-02-23
Integration tests - Starting to share with custom groups

SergioBertolinSG:
@PVince81 The starting test case is failing, something not working or the test is not correct?.


tests/integration/custom_groups_features/Sharing-CustomGroups.feature
```diff
@@ -0,0 +1,20 @@
...
+
+Scenario: Creating a share with a group
+	Given As an "admin"
+	And user "user0" exists
+	And user "user1" exists
+	And user "member1" exists
+	And user "user1" created a custom group called "sharing-group"
+	And user "user1" made user "member1" member of custom group "sharing-group"
+    And As an "user0"
+    When sending "POST" to "/apps/files_sharing/api/v1/shares" with
```
PVince81: Why not simply `And file "welcome.txt" of user "user0" is shared with group "sharing-group"` ?

PVince81:
Maybe the file "welcome.txt" doesn't exist ? Are skeleton files copied when running integration tests from apps ?

SergioBertolinSG:
Ah right, they don't have the skeleton.

Missing this
#Set up personalized skeleton
$OCC config:system:set skeletondirectory --value="$(pwd)/skeleton"****

tests/integration/custom_groups_features/Sharing-CustomGroups.feature
SergioBertolinSG: Because it is the simplest example, I like to have the simplest as a call to the api and make that abstract in the next test.

SergioBertolinSG:
-    1 => 'sharing-group'
      +    1 => 'customgroup_17'

Problem with the group names. They have as name customgroup_X, being X an autoincremental value.

We should avoid this. @PVince81 is this a bug in the app? 

Not sure if this is the displayname, but if that is the case, should we set a display name after creation?

PVince81:
Ah yes, this is by design. I needed an internal group id that doesn't conflict with any other user backends so it's "customgroups_" + numeric id of the custom group.

The display name is used in the sharing dialog.

Now considering that the "uri" of a custom group is supposed to be unique, I could replace the autoincrement value with the uri. So the id would become "customgroups_sharing-group".

Note that the "uri" is specified at creation time and can never change even on rename (like for CalDAV).

Would that make sense ?

SergioBertolinSG:
If two users want to have each a custom group called "somestring", they won't be able to do that with this approach? the uri for the custom group of the first user would be "customgroups_somestring" but what about the second?

PVince81:
The client needs to contain logic to check for duplicates.
If you do a mkcol on "somestring" and you get 405, then you need to try "somestring2".
But in the end you still set the display name to "somestring".

I had to implement that logic into the frontend code... Webdav fun...

PVince81:
Now using uri field for compoung group backend id https://github.com/owncloud/customgroups/pull/26/commits/ba84877abd61d9a0cd5c694f8b9329e4a5c055f6 (@jvillafanez @DeepDiver1975 FYI)

The test now passes locally.

PVince81:
@jvillafanez @DeepDiver1975 please review this change

jvillafanez:
Code looks good so far.

PVince81:
looks like I missed some stuff.

PVince81:
Rebased to get the pagination changes in as well.

@SergioBertolinSG if all tests pass now I suggest we merge this before continuing. I want to have the uri stuff on master.

PVince81:
Fixed role filter calls

PVince81:
Merged now. @SergioBertolinSG please do a separate PR for your new tests. Let's hope you won't find any new issues 😉

Issue #16, opened 2017-02-06
Pagination with REPORT


PVince81:
Currently we use PROPFIND for many listing operations.

We need to also provide a REPORT method that supports pagination for these operarations.

PVince81:
We'll also need to add filtering to be able to add search boxes in the UI for convenience.

Pagination is an optimization for scalabilty but alone it doesn't make much sense if we don't add quicksearch too.

@DeepDiver1975

PVince81:
Pagination here: https://github.com/owncloud/customgroups/pull/27

PR #27, opened 2017-02-23
Add pagination/search through REPORT

PVince81:
Fixes https://github.com/owncloud/customgroups/issues/16

Example report:
```xml
<?xml version="1.0" encoding="utf-8" ?>
<oc:search-query xmlns:a="DAV:" xmlns:oc="http://owncloud.org/ns">
        <a:prop>
                <oc:role></oc:role>
                <oc:display-name></oc:display-name>
        </a:prop>
        <oc:search>
                <oc:pattern>us</oc:pattern>
                <oc:limit>5</oc:limit>
                <oc:offset>0</oc:offset>
        </oc:search>
</oc:search-query>
```

You can apply this to the following URLs:
- groups list: `curl -i --data-binary "@report-customgroups.xml" -X REPORT -H "Content-Type: text/xml" http://admin:admin@localhost/owncloud/remote.php/dav/customgroups/groups`
- group members list: `curl -i --data-binary "@report-customgroups.xml" -X REPORT -H "Content-Type: text/xml" http://admin:admin@localhost/owncloud/remote.php/dav/customgroups/groups/group1`
- user membership list: `curl -i --data-binary "@report-customgroups.xml" -X REPORT -H "Content-Type: text/xml" http://admin:admin@localhost/owncloud/remote.php/dav/customgroups/users/user1/`

@DeepDiver1975 @jvillafanez please review.

If the implementation style is ok I'll add matching unit tests.

- [ ] unit tests

### DIFF
--- a/lib/CustomGroupsBackend.php
+++ b/lib/CustomGroupsBackend.php
@@ -79,7 +79,7 @@ class CustomGroupsBackend implements \OCP\GroupInterface {
 	 * @return array an array of group names
 	 */
 	public function getUserGroups($uid) {
-		$memberInfos = $this->handler->getUserMemberships($uid, null);
+		$memberInfos = $this->handler->getUserMemberships($uid, null, null);
 		return array_map(function ($memberInfo) {
 			return $this->formatGroupId($memberInfo['group_id']);
 		}, $memberInfos);
@@ -96,7 +96,7 @@ class CustomGroupsBackend implements \OCP\GroupInterface {
 	 * @return array an array of group names
 	 */
 	public function getGroups($search = '', $limit = -1, $offset = 0) {
-		$groups = $this->handler->searchGroups($search, $limit, $offset);
+		$groups = $this->handler->searchGroups(new Search($search, $offset, $limit));
 		return array_map(function ($groupInfo) {
 			return $this->formatGroupId($groupInfo['group_id']);
 		}, $groups);
--- a/lib/CustomGroupsBackend.php
+++ b/lib/CustomGroupsBackend.php
@@ -79,7 +79,7 @@ class CustomGroupsBackend implements \OCP\GroupInterface {
 	 * @return array an array of group names
 	 */
 	public function getUserGroups($uid) {
-		$memberInfos = $this->handler->getUserMemberships($uid, null, null);
+		$memberInfos = $this->handler->getUserMemberships($uid, null);
 		return array_map(function ($memberInfo) {
 			return $this->formatGroupId($memberInfo['group_id']);
 		}, $memberInfos);
--- a/lib/CustomGroupsDatabaseHandler.php
+++ b/lib/CustomGroupsDatabaseHandler.php
@@ -85,12 +85,11 @@ class CustomGroupsDatabaseHandler {
 	 * Get all group memberships of the given user
 	 *
 	 * @param string $uid Name of the user
-	 * @param null|int $roleFilter optional role filter
 	 * @param Search $search search
 	 * @return array an array of member info
 	 * @throws \Doctrine\DBAL\Exception\DriverException in case of database exception
 	 */
-	public function getUserMemberships($uid, $roleFilter = null, $search = null) {
+	public function getUserMemberships($uid, $search = null) {
 		$qb = $this->dbConn->getQueryBuilder();
 		$qb->select('m.group_id', 'm.user_id', 'm.role', 'g.uri', 'g.display_name')
 			->from('custom_group_member', 'm')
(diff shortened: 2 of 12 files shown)


PVince81:
Do we agree on the general approach so I can solidify this with unit tests ?

lib/CustomGroupsDatabaseHandler.php
```diff
@@ -409,4 +408,30 @@ private function formatMemberInfo(array $row) {
...
+	 * @param IQueryBuilder $qb query builder
+	 * @param Search $search search
+	 * @param string $property property to apply search on
+	 * @return IQueryBuilder query builder
+	 */
+	private function applySearch(IQueryBuilder $qb, Search $search, $property = null) {
+		if ($search !== null) {
+			if ($search->getPattern() !== null && $property !== null) {
+				$likeString = '%' . $this->dbConn->escapeLikeParameter(strtolower($search->getPattern())) . '%';
+				$qb->andWhere($qb->expr()->like($qb->createFunction('LOWER(`' . $property . '`)'), $qb->createNamedParameter($likeString)));
```
jvillafanez: that `lower` function is strictly needed?. Although it might not be risky, the behaviour isn't optional and isn't documented.

lib/CustomGroupsDatabaseHandler.php
```diff
@@ -85,10 +86,11 @@ public function inGroup($uid, $numericGroupId) {
 	 *
 	 * @param string $uid Name of the user
 	 * @param null|int $roleFilter optional role filter
+	 * @param Search $search search
 	 * @return array an array of member info
 	 * @throws \Doctrine\DBAL\Exception\DriverException in case of database exception
 	 */
-	public function getUserMemberships($uid, $roleFilter = null) {
+	public function getUserMemberships($uid, $roleFilter = null, Search $search = null) {
```
jvillafanez: I'm not sure if we can switch the `roleFilter` and `search` parameters at this point. I've seen too much `getUserMembership(uid, null, search)`. A `getUserMembership(uid, search)` with the `roleFilter` as null by default looks better.

jvillafanez:
The approach looks good to me.

lib/CustomGroupsDatabaseHandler.php
PVince81: I also hesitated to maybe put the role filter as a member of the `Search` class, what do you think ?

lib/CustomGroupsDatabaseHandler.php
PVince81: That's how we did search in the default group backend. I think it makes sense to keep it.

PVince81:
@jvillafanez thanks a lot for your feedback !

I have the feeling that having a `Search` object is the way to go in the future for any paginated queries as it saves having a lot of arguments.

PVince81:
- [ ] fix DB handler unit tests
- [ ] add tests for REPORT
- [ ] include `roleFilter` in the `Search` class and REPORT args ?

PVince81:
- [ ] unit tests for `->search()` method on nodes

PVince81:
Ok, I'll try adding `roleFilter` into the `Search` object. It should work in most queries.

PVince81:
This is ready for review now

PVince81:
Unit tests added + a few tweaks

lib/CustomGroupsBackend.php
```diff
@@ -79,7 +79,7 @@ public function inGroup($uid, $gid) {
 	 * @return array an array of group names
 	 */
 	public function getUserGroups($uid) {
-		$memberInfos = $this->handler->getUserMemberships($uid, null);
+		$memberInfos = $this->handler->getUserMemberships($uid, null, null);
```
jvillafanez: too many nulls here

lib/Dav/GroupMembershipCollection.php
```diff
@@ -214,12 +215,16 @@ public function getChild($userId) {
...
+	}
+
+	public function search(Search $search = null) {
 		$groupId = $this->groupInfo['group_id'];
 		if (!$this->helper->isUserMember($groupId)
 			&& !$this->helper->isUserAdmin($groupId)) {
 			throw new Forbidden("No permission to list members of group \"$groupId\"");
 		}
-		$members = $this->groupsHandler->getGroupMembers($groupId);
+		$members = $this->groupsHandler->getGroupMembers($groupId, null, $search);
```
jvillafanez: are the parameters correct?

PVince81: uh oh... why didn't the tests explode... I'll recheck

PVince81:
@jvillafanez fixed now...

The reason why tests passed is because the expectation was also wrong...

PVince81:
and now Travis is imitating Jenkins it seems. It's been an hour and I still can't see any results (loading forever)

PVince81:
```
    /home/travis/build/owncloud/core/apps/customgroups/tests/integration/custom_groups_features/CustomGroups.feature:147

    /home/travis/build/owncloud/core/apps/customgroups/tests/integration/custom_groups_features/CustomGroups.feature:159
```

PVince81:
Going to merge this anyway now and resolve conflicts and integration tests as part of https://github.com/owncloud/customgroups/pull/26

Issue #4, opened 2016-12-03
Send notification when added to group


PVince81:
Whenever a group admin adds a user to a group, send that user a notification using the notification API.

(note that as of today notifications do not send email but appear only in the web UI and clients that support notifications)

PVince81:
PR is here: https://github.com/owncloud/customgroups/pull/28

PR #28, opened 2017-02-24
Publish notification when adding user

PVince81:
Fixes https://github.com/owncloud/customgroups/issues/4

- [x] Will require https://github.com/owncloud/customgroups/pull/13 to be able to fix the links.

### DIFF
--- a/appinfo/app.php
+++ b/appinfo/app.php
@@ -21,3 +21,4 @@
 
 $app = new \OCA\CustomGroups\Application();
 $app->registerGroupBackend();
+$app->registerNotifier();
--- a/tests/unit/Dav/GroupsCollectionTest.php
+++ b/tests/unit/Dav/GroupsCollectionTest.php
@@ -29,6 +29,8 @@ use OCA\CustomGroups\Dav\GroupMembershipCollection;
 use OCA\CustomGroups\Service\MembershipHelper;
 use OCP\IGroupManager;
 use OCA\CustomGroups\Search;
+use OCP\IURLGenerator;
+use OCP\Notification\IManager;
 
 /**
  * Class GroupsCollectionTest
@@ -78,7 +80,9 @@ class GroupsCollectionTest extends \Test\TestCase {
 			$this->handler,
 			$this->userSession,
 			$this->userManager,
-			$this->groupManager
+			$this->groupManager,
+			$this->createMock(IManager::class),
+			$this->createMock(IURLGenerator::class)
 		);
 		$this->collection = new GroupsCollection($this->handler, $this->helper);
 	}
--- a/tests/unit/NotifierTest.php
+++ b/tests/unit/NotifierTest.php
@@ -1,0 +1,77 @@
+<?php
+/**
+ * @author Vincent Petry <pvince81@owncloud.com>
+ *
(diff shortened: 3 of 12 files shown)


PVince81:
- [ ] discard/kill that notification if the user membership was removed before the notification is shown or the group was deleted

PVince81:
- [ ] BUG: router doesn't return correct link

PVince81:
Ready for review @jvillafanez @DeepDiver1975 @IljaN @VicDeo 

I'll investigate how to kill the notification in case of deletion separately, seems to be tricky.

appinfo/app.php
```diff
@@ -21,3 +21,4 @@
 
 $app = new \OCA\CustomGroups\Application();
 $app->registerGroupBackend();
+$app->registerNotifier();
```
DeepDiver1975: cries for info.xml enhancement ....

tests/unit/Dav/GroupsCollectionTest.php
```diff
@@ -78,7 +79,9 @@ public function setUp() {
 			$this->handler,
 			$this->userSession,
 			$this->userManager,
-			$this->groupManager
+			$this->groupManager,
+			$this->createMock(\OCP\Notification\IManager::class),
```
DeepDiver1975: use ?

tests/unit/NotifierTest.php
```diff
@@ -0,0 +1,76 @@
...
+	 * @var Notifier
+	 */
+	private $notifier;
+
+	/**
+	 * @var CustomGroupsDatabaseHandler
+	 */
+	private $handler;
+
+	public function setUp() {
```
DeepDiver1975: parent:setup?

tests/unit/NotifierTest.php
```diff
@@ -0,0 +1,76 @@
...
+
+		$this->handler = $this->createMock(CustomGroupsDatabaseHandler::class);
+
+		$this->notifier = new Notifier(
+			\OC::$server->getL10NFactory(),
```
DeepDiver1975: mock?

tests/unit/Dav/GroupsCollectionTest.php
PVince81: I could... but I somehow hate having `IManager` without knowing what a "manager" is.
This should have been called `INotificationManager` in the first place.

Or I could use `use xxx as yyy` ?

tests/unit/NotifierTest.php
PVince81: I don't want to mock it... too many calls to it and in the end the default one works fine as it will return the same texts untranslated but formatted

appinfo/app.php
PVince81: if we had a generic service registration we wouldn't need to enhance core for everything out there

`<services><service type="\OC\INotifier">\OCA\CustomGroups\Notifier</service></services>`

then core (or any app) could just find all services of a given type/namespace.

That was what I hoped to achieve with https://central.owncloud.org/t/service-based-app-imlpementation-to-solve-inter-app-dependencies/1615

PVince81:
@DeepDiver1975 fixed the namespace use and added `parent::setUp`.

Other pieces aren't blockers from my POV.

Issue #6, opened 2016-12-05
Hide custom groups from admin / users page


PVince81:

PVince81:
- [ ] hide from admin
- [ ] hide from Webdav principals

PVince81:
I tried several approaches, but none is satisfactory.

One of the approaches adds a new method param `$manageableOnly` to group-related methods on the IGroupManager. The trouble with this is that the group manager caches groups, so it would need to cache them both, one with `$manageableOnly = true` and one with false...

Now I could first retrieve all groups and then filter them later in the UsersController or Group\Metadata which is used to populate the UI...

PVince81:
Tentative PR here: https://github.com/owncloud/core/pull/27184

PVince81:
or maybe we should just allow them to appear there...

The one problem I see is that they are not distinguishable from regular groups. One idea would be to introduce the backend name, so it could appear as:
- localgroup (Local)
- ldapgroup (LDAP)
- My custom group (Custom group)

For this the backend id needs to be exposed to the UI code with a friendly name.

But this brings other issues: the list of groups could become long and it would need to be searchable.

That or simply have a checkbox for the admin "don't show groups from these backends: "... That would be close to what this PR does.

However the groups would still appear in the provisioning API, and there there is also no information about what backend they come from...

PVince81:
- Approach 1: post-filter group list: https://github.com/owncloud/core/pull/27184
- Approach 2: `getUserGroupsForSharing`, sharing specific APIs: https://github.com/owncloud/customgroups/pull/25
- Approach 3: actually leave them visible but use their display names: https://github.com/owncloud/core/pull/27186

jvillafanez:
I think the group backend interface should include a `isVisibleFor($subsystem, $user)`. Basically core would ask the backend if that group is visible for the subsystem (sharing, webdav, user listing, etc) and the current user.

The advantages I see with that approach:
* Backends have control over their own visibility in well-known core components. It's their choice to show or hide their groups.
* This is extensible to 3rdparty apps as long as both apps know each other. 3rdparty can check if the group backend is visible for "my_custom_app" and decide.

PVince81:
Hmm, this might work I guess, I'll give it a try.

I didn't like the idea of having a backend knowing about UI stuff, but there might be no other way.

For now these are the subsystems I'd list:
- "management": meaning users page for management and also provisioning API (no distinction between the two, the users page should really use the provisioning API...)
[...]

PR #30, opened 2017-03-01
Filter by scope

PVince81:
Fixes https://github.com/owncloud/customgroups/issues/6

Requires https://github.com/owncloud/core/pull/27287

### DIFF
--- a/lib/CustomGroupsBackend.php
+++ b/lib/CustomGroupsBackend.php
@@ -180,4 +180,14 @@ class CustomGroupsBackend implements \OCP\GroupInterface {
 	private function formatGroupId($uri) {
 		return self::GROUP_ID_PREFIX . $uri;
 	}
+
+	/**
+	 * Returns true only if the scope is "sharing"
+	 *
+	 * @param string $scope
+	 * @return bool true if the scope is "sharing"
+	 */
+	public function isVisibleForScope($scope) {
+		return ($scope === 'sharing');
+	}
 }


jvillafanez:
👍 

Question: groups will be hidden if the scope isn't "sharing", do we want that?
* Show customs groups only for sharing  (current)
* Hide customs groups from management / user listing. Custom groups might be visible in other scopes.

I think this is fine, so no need to change anything.

PVince81:
So far I didn't try to find all code paths that could pass a scope but didn't.

Regarding custom groups, we only ever want to show them for sharing, nowhere else. So if there's another "empty scope" case somewhere out there, we don't show them there.

More possible scopes that come to mind:
- autocomplete fields to disable apps for specific groups: let's exclude custom groups there
- autocomplete fields for admins to set external storage "applicable" field: let's exclude too
- autocomplete fields for "Exclude these groups from sharing": also exclude custom groups there
- more autocomplete fields related to sharing in the admin page: I'd also exclude custom groups from now.

Now if later we say we do want custom groups there, we could add a scope there later.

jvillafanez:
This is more a whitelist vs blacklist topic. A whitelist makes more sense in this particular case.

If core enables more scopes, it's fine to add them in the whitelist later.

PVince81:
Sounds good, whitelist it is then!

PVince81:
Rebased to pick up the changes merged in core.

PR #31, opened 2017-03-06
Changed paths to new location

SergioBertolinSG:


DeepDiver1975:
:+1:

SergioBertolinSG:
The reason behind last commit is that sabre dav doesn't throw a ClientHttpException in the failure and ClientException doesn't include response.

```
No, that is not available Fatal error: Call to undefined method Sabre\HTTP\ClientException::getResponse() (Behat\Testwork\Call\Exception\FatalThrowableError)
```
```php
if ($response->getStatus() === 207) {
           // If it's a 207, the request could still have failed, but the
           // information is hidden in the response body.
````

it returns 207

Issue #37, opened 2017-03-09
Makefile must sign app


PVince81:
Make it possible to specify a key in an env variable and use that key with occ to sign the app.

We'll likely need such process to make it easier to release signed market apps.

CC @DeepDiver1975 @VicDeo

DeepDiver1975:
Contacts and twofactor_totp use makefile to sign - https://github.com/owncloud/contacts/blob/master/Makefile#L52-L63

DeepDiver1975:
feel free to adopt

PR #39, opened 2017-03-09
Added signing rules

PVince81:
Fixes https://github.com/owncloud/customgroups/issues/37

It will sign the app before building the tarball, if possible.

Signing worked but when I put the app on git/core it still shows as experimentall.

@DeepDiver1975 @jvillafanez


jvillafanez:
👍

PVince81:
looks like I broke the js test deps, fixing...

Issue #32, opened 2017-03-08
Autocomplete for the member field


PVince81:
Use the same API like sharing so we benefit from "restrict sharing to only your groups"

PVince81:
This is critical especially for LDAP cases where the user id is not user friendly.

PVince81:
Some challenges:

- "restrict groups to only share with groups they are member of"
- when user autocomplete is disabled, only allow full user name
- filtering out sharing results on the server side (client side is not possible if we use pagination as we don't have all results)
    - this makes it impossible to use the existing sharees API and requires reimplementation in the custom groups code

PVince81:
and for filtering on the server-side, best would be to be able to JOIN between the user account table and members table in a way that excludes already existing members. But that requires the user account table to exist. Since this would join with the customgroups specific table, this code needs to live in this app.

The other alternative, less nice is to add an argument in core for the search:
`UserManager->search(..., array $excludeResults)` where `$excludeResults` is an array of all user ids that already exist and that need to be filtered out.

Or yet another method: implement an autocomplete controller in customgroups (private API, app framework) and have that one first call `UserManager->search()` and then filter out the existing members from the results. This does not work well with pagination because maybe the first page contains members that were already added, so this would cause an empty result set after filtering. This means that this code would need to go forward page by page until we have enough results to output.

And besides this, post-filtering might not be very efficient. Best would be DB-level join because we can also paginate on that...

PVince81:
Summarizing what I posted in the chat:

I want to build the customgrous autocomplete which looks like the share panel autocomplete when sharing, this to select members to add to a group
now I need to filter out the members that are already in the group from the autocomplete box
the way how the sharing dialog did was filter out the results on the JS side based on what's on the list already
this sucks because it's not compatible with pagination (we never have the full list on JS side)
possible approaches:
1) ignore the pagination issue (if we implement it later) and filter on the JS side: makes it possible to reuse the "sharees" endpoint like the share dialog. I have a prototype of this that works.
2) create a new endpoint in customgroups for the autocomplete results and have the "filter users already in the list" on the PHP side. For that there are at least two approaches:
a) UserManager->search(...) with pattern and then post-filter with a select on the custom group members table: sucks because it's a PHP level join-like construct
b) extend UserManager->search(...) to accept a list of exclusions. But also needs a select on the custom group members table to pass it there. Same suckage as a)
[...]

PR #43, opened 2017-03-14
Member autocomplete

PVince81:
Fixes https://github.com/owncloud/customgroups/issues/32


PVince81:
TODO:

- [ ] squash on merge (I kept the older commits in case we do want select2 or something else...)
- [ ] add unit tests for the controller paging logic

PVince81:
This ready for review. Don't be scared by the number of files changes, I moved one class into a different namespace so had to adjust the `use` statements in a lot of files.

Please review @jvillafanez @DeepDiver1975 @VicDeo

js/MembersInputView.js
```diff
@@ -0,0 +1,163 @@
...
+				data: {
+					group: this.groupUri,
+					pattern: search.term.trim(),
+					limit: 200
+				}
+			}).done(function (result) {
+					$loading.addClass('hidden');
+					$loading.removeClass('inlineblock');
+
+					if (result.message) {
```
jvillafanez: Maybe we should be more strict with this check. Just checking whether there is a message to show an error...

js/MembersInputView.js
```diff
@@ -0,0 +1,163 @@
...
+						OC.Notification.showTemporary(result.message);
+						response();
+						return;
+					}
+
+					result = result.results;
```
jvillafanez: Risky naming here. I'd use a new variable to avoid confusion.

js/MembersInputView.js
```diff
@@ -0,0 +1,163 @@
...
+			var text = item.displayName;
+			var insert = $("<div class='customgroups-autocomplete-item'/>");
+			var avatar = $("<div class='avatardiv'></div>").appendTo(insert);
+			avatar.avatar(item.userId, 32, undefined, undefined, undefined, item.displayName);
+
+			$("<div class='autocomplete-item-text'></div>")
+				.text(text)
+				.appendTo(insert);
+			insert.attr('title', item.userId);
+			insert = $("<a>")
```
jvillafanez: Use a new variable?

lib/Controller/PageController.php
```diff
@@ -37,4 +64,31 @@ public function index() {
...
+
+		$results = $this->helper->searchForNewMembers($groupInfo['group_id'], $pattern, $limit);
+		$results = array_map(function (IUser $entry) {
+			return [
+				'userId' => $entry->getUID(),
+				'displayName' => $entry->getDisplayName()
+			];
+		}, $results);
+
+		return new DataResponse(['results' => array_values($results)], Http::STATUS_OK);
```
jvillafanez: I guess we can leave it as a technical debt for now, but we're traversing the list two more times than we ideally should (one for the array_map and another for the array_values).
I'm not sure if it's easy to improve, but anyway I don't expect a list huge enough to have a performance impact.

lib/Service/MembershipHelper.php
```diff
@@ -59,7 +61,7 @@ class MembershipHelper {
 	 *
 	 * @var IGroupManager
 	 */
-	private $groupsManager;
+	private $groupManager;
```
jvillafanez: ❗️ Any reason for this change? The new code doesn't use this, so I guess the old code is wrong?

lib/Service/MembershipHelper.php
```diff
@@ -177,4 +179,58 @@ public function isTheOnlyAdmin($groupId, $userId) {
...
+	 * This searches the whole user list by display name and excludes the
+	 * users that are already members of the given group.
+	 *
+	 * @param int $groupId numeric group id for which to find new members
+	 * @param string $pattern pattern to search for in display names
+	 * @param int $limit limit up to which to return results
+	 * @return IUser[] results
+	 */
+	public function searchForNewMembers($groupId, $pattern, $limit = 200) {
+		$existingMembers = $this->getGroupMemberUserIds($groupId, $pattern);
```
jvillafanez: I wonder if there isn't any alternative other than fetch all the group members at once. I think we can apply the same reason as above and say that we don't expect a lot of members inside a group, but it's something we should check at some point.

lib/Service/MembershipHelper.php
PVince81: Yes, it was wrong before. I happened to notice the phpmd error telling me it was unused so I fixed it.

lib/Service/MembershipHelper.php
PVince81: We don't fetch all group members at once: the $pattern is passed so this should at least limit the number of results a bit to something that makes sense to join with the other results.

js/MembersInputView.js
```diff
@@ -0,0 +1,163 @@
...
+		_rendered: false,
+
+		_newTag: null,
+
+		className: 'member-input-view',
+
+		template: function() {
+			var $el = $('<div></div>');
+			$el.append($('<input class="member-input-field" type="text"/>')
+				.attr('placeholder', t('customgroups', 'Add user to this group')));
```
PVince81: Will rather adjust to make it clearer

js/MembersInputView.js
PVince81: I'll check the xhr status, if it's 200 then continue. Else check result.message else show generic message.

js/MembersInputView.js
PVince81: will do

js/MembersInputView.js
PVince81: I definitely didn't write this code, was copy-pasted from the share dialog autocomplete...

Will beautify it a bit.

lib/Controller/PageController.php
PVince81: Good catch. I was able to get rid of `array_values` completely.

It was some remains from a previous attempt where I had `array_filter`, because `array_filter` filters out values from an array but doesn't reaaranges the indices... So serializing an array to json when the array has no continuous indices results in a hash JSON array instead of a real JSON array...

PVince81:
@jvillafanez thanks a lot for your review. I have addressed all comments either with code or answers 😄

js/MembersInputView.js
```diff
@@ -0,0 +1,167 @@
...
+		autocompleteRenderItem: function($ul, item) {
+			var text = item.displayName;
+			var $item = $("<div class='customgroups-autocomplete-item'/>");
+			var $avatar = $("<div class='avatardiv'></div>").appendTo($item);
+			$avatar.avatar(item.userId, 32, undefined, undefined, undefined, item.displayName);
+
+			$("<div class='autocomplete-item-text'></div>")
+				.text(text)
+				.appendTo($item);
+			$item.attr('title', item.userId);
```
jvillafanez: I think we should make this function more readable. Right now it's difficult to read due to context switches: create an element, do something to another, change back to the first; append vs appendTo in the same function...

I'd rather create and set all the elements we need one by one and then insert them from botton to top.
```
var a = $("whatever").attr(...).text(....)
var b = $("w2").text(....)
var c = $("w3").attr(....)
c.appendTo(b);
b.appentTo(a);
```

PVince81: Ok you know what ? Let's do this cleanly and use an actual handlebars template.

PVince81:
@jvillafanez moved to templates, wayyyy more readable. 😄

jvillafanez:
👍 x2

Issue #23, opened 2017-02-20
Show user display names instead of user id


PVince81:
Requires an API in core because we don't want to fetch every user's display name one by one...

PVince81:
- [ ] REQUIRES: https://github.com/owncloud/core/issues/27235

PVince81:
Should be possible to start implement this but with a shim.
Goal is to have the code ready for when the API itself is ready to be plugged in.

PVince81:
and also show avatars

PVince81:
This is critical especially for LDAP cases where the user id is not user friendly.

PVince81:
PR here with a barbaric and inefficient method: https://github.com/owncloud/customgroups/pull/45

PVince81:
Let's leave the ticket open for the non-barbaric approach

PR #45, opened 2017-03-14
Return member display name when queried

PVince81:
Uses a barbaric method to retrieve the member display name: basically get the user from the user manager and call `$user->getDisplayName()`. Internally this will likely cause multiple queries like "does the user exist" and also "get the display name of that specific user".
So when dealing with a list of 20 entries there will be 20+ queries.

This should be changed once the user account table is ready and we might be able to join against it.

The current approach was done only for feature completion.

Please review @jvillafanez @DeepDiver1975

### DIFF
--- a/js/MemberModel.js
+++ b/js/MemberModel.js
@@ -30,7 +30,8 @@
 		},
 
 		davProperties: {
-			'role': NS + 'role'
+			'role': NS + 'role',
+			'userDisplayName': NS + 'user-display-name'
 		}
 	});
 
--- a/js/MembersView.js
+++ b/js/MembersView.js
@@ -189,7 +189,7 @@
 
 			// TODO: use displayName once available
 			OC.dialogs.confirm(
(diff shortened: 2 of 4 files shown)


PVince81:
Fixes https://github.com/owncloud/customgroups/issues/23

One could argue that if we're returning paginated results anyway, the number of extra queries will always be limited to the number of results, so it would not go too crazy with 1000+ results.

jvillafanez:
As long as we know the current limitations is fine for me. 👍 

Maybe we should document the recommended limits. If we suppose the performance drop with 50 members, this would be acceptable for usual cases, so we could deliver this solution until we find a better one. On the other hand, if the performance drops with 10 user this could be unusable for the average user.

PVince81:
In general I think this will stop working as soon as we think about sorting... Because sorting on display names that are retrieved like this is shit.

But I want to have this hack merged first so we can at least demo something.
And then once the account table exists, rework this bit. The idea will be that the "CustomGroupsDatabaseHandler" will join on the account table and have the user's display name already available in the returned result.

Issue #50, opened 2017-03-27
Superadmin cannot add users to groups


PVince81:
Whenever a superadmin (aka ownCloud admin) is not a group member, they should still be able to add members to existing groups.

Currently it fails on the backend.

PVince81:
@SergioBertolinSG wasn't this covered by the integration tests ?

SergioBertolinSG:
Yes, IIRC it worked well.

SergioBertolinSG:
Ok, this specific case is not covered, I'll add it.

PR #51, opened 2017-04-03
Added tests for superadmin

SergioBertolinSG:
To cover https://github.com/owncloud/customgroups/issues/50


PVince81:
Fixed now.

The `childExists` method was only checking for membership but didn't have a special case for super-admins while `getChildren()` did

Issue #53, opened 2017-04-07
Improve notifications


PVince81:
- Notifications
  - Create a group “qustom” as user “admin”, add user “test3” to group and promote to group admin
  - See notification below
    - Notification for being added should say “You have been added to group ‘qustom’ by ‘admin’.
    - Notification for promotion should say: “Admin” assigned ‘Group admin’ role for the group “qustom” to you.
    - For making a group admin a member again there is no notification → consistency ("Admin" assigned ‘Member’ role for the group “qustom” to you.)
![custom_groups_notification2](https://cloud.githubusercontent.com/assets/21193500/24759933/55106752-1ae7-11e7-84c4-e63d8fe5bdc5.png)

@pmaier1

PVince81:
I'll also add a message about being removed from a group.

PVince81:
PR here https://github.com/owncloud/customgroups/pull/56

pmaier1:
One thing I just discovered that is missing and could maybe be added: Notification for group deletion

PVince81:
Who would receive this notification ? All members ?

pmaier1:
Yes, I would say so

PVince81:
Okay, seems we might need to use a different approach here.

So far we sent the notification directly at the time of action.

Here since we need to notify all members (which could be quite a lot), we might want to move this to a background job. If we don't, then the delete operation will take much longer for the user (spinner for a longer time).

pmaier1:
> Okay, seems we might need to use a different approach here.
So far we sent the notification directly at the time of action.
Here since we need to notify all members (which could be quite a lot), we might want to move this to a background job. If we don't, then the delete operation will take much longer for the user (spinner for a longer time).

Yep, this definitely sounds like the right approach! It's not a blocker but should be added for completeness and consistency.

PVince81:
Raised here https://github.com/owncloud/customgroups/issues/57.
The background job approach will introduce a bit more complexity.

PR #56, opened 2017-04-10
Improved notifications

PVince81:
Added notification for removal of membership.
Added notification for role changes.
Simplified notification subject line.
Added unit tests for notification sending.

Fixes https://github.com/owncloud/customgroups/issues/53

### DIFF
--- a/lib/Notifier.php
+++ b/lib/Notifier.php
@@ -26,6 +26,8 @@ use OCP\Notification\IManager;
 use OCP\Notification\INotification;
 use OCP\Notification\INotifier;
 use OCA\CustomGroups\CustomGroupsDatabaseHandler;
+use OCA\CustomGroups\Dav\Roles;
+use OCP\IL10N;
 
 class Notifier implements INotifier {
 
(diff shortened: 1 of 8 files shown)


PVince81:
@pmaier1 if you want to review wording, start reading from here: https://github.com/owncloud/customgroups/pull/56/files#diff-58b7c72093e85043455e237d9a66dc1eR70

"Subject" is the line you saw in bold in https://github.com/owncloud/notifications/issues/93 and "Message" is the part below.

pmaier1:
As far as I understand the code (^^) it looks right but in the screenshot https://github.com/owncloud/notifications/pull/93#issue-220590221 it looks like the variables $roleName and $group would be mixed up for the subject when roles are changed (like 'Role change in group $roleName').

lib/Notifier.php
```diff
@@ -65,16 +67,57 @@ public function prepare(INotification $notification, $languageCode) {
...
+			if ($notification->getMessage() === 'removed_member') {
+				$notification->setParsedMessage(
+					$l->t('You have been removed from the group "%2$s" by "%1$s".', $notification->getMessageParameters())
+				);
+			}
+
+			if ($notification->getSubject() === 'changed_member_role') {
+				$groupName = $notification->getSubjectParameters()[1];
+				$notification->setParsedSubject(
+					$l->t('Role change in group "%1$s".', [$groupName])
```
pmaier1: the variable needs to be called '$group', right?

PVince81: no, the parameter in the middle has a special syntax. The "1" inside means the first parameter from the array you see behind: since the array `[$groupName]` only has one value it will be that one.

PVince81: hmmm, you're right. Actually $groupName reads the wrong subject parameter... that's yet another array. I'll check the position

PVince81: I just retested locally and all is correct. It is likely that I made that screenshot earlier before I fixed the order.

pmaier1: ok, nice :) :+1:

PR #63, opened 2017-05-11
Fix integration tests with missing mailhog variable.

SergioBertolinSG:
Any change in core affects these tests.

### DIFF
--- a/tests/integration/config/behat.yml
+++ b/tests/integration/config/behat.yml
@@ -16,6 +16,7 @@ default:
               - admin
               - admin
             regular_user_password: 123456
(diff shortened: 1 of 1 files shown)


PVince81:
it's a bit sad that every app even the ones that don't use this will need adjustments because of core.

👍 for now

SergioBertolinSG:
I know, but when using mailhog you need the url to connect and that should be in the config file.

If you think about a better solution, we can discuss it afterwards.

PVince81:
if there was a way to make that variable option it would be nice, not sure if behat supports that.

then when actually connecting to mailhog if the var is not set, fail

PR #68, opened 2017-05-30
Fix when avatars are disabled

### LINKED ISSUE #60: Settings page broken when avatars are disabled
1. Set "enable_avatars" to false in config.php
1. Open settings page
1. Create a group

Expected: still works
Actual: broken with JS errors.

Maybe need some shims... I'm pretty sure this is not the only app that is likely to break.

PVince81:
Fixes #60

@ppaysant please review


tomneedham:
:+1 tested and works

PR #73, opened 2017-06-26
Implement method that returns users in group

PVince81:
Required for some apps like activity which use this to find out what
group members there are to send emails to. This could have other side
effects.

### DIFF
--- a/lib/CustomGroupsBackend.php
+++ b/lib/CustomGroupsBackend.php
@@ -135,7 +135,7 @@ class CustomGroupsBackend implements \OCP\GroupInterface {
 	}
 
 	/**
-	 * Not supported, returns an empty array.
+	 * Returns all users in a custom group.
 	 *
 	 * @param string $gid group id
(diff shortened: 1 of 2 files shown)


PVince81:
- [ ] check for potential side effects
- [ ] add unit test

PVince81:
Added unit test

PVince81:
I grepped the whole core code for `searchUsers(` with a few apps and found that this would actually fix another bug with sharing: when sharing with a user or group, there is this checkbox to "inform recipients". Apparently the code for this is using `$group->searchUsers()` to find who to inform. This means that this PR also fixed this bug.

The other two occurrences were in the activity app, which is the one this PR is for.

@jvillafanez can you review this ?

jvillafanez:
I'd add some log to help to debug, but other than that, code looks fine 👍

PVince81:
Not too convinced about having debug log in this code path. Anyway, worst case we can quickly add them and do another release.

Issue #77, opened 2017-06-29
Add (2) when creating duplicate group name


PVince81:
Whenever a user creates a new group, automatically append (2) or another number if the group list visible for this user already contains a similar entry. This is mostly for information purposes.
The user can still choose to rename the group back to the name.

Should not append anything if the group exists but is not visible for the current user.

Estimate: 1-3 days as we might have to tear up the server side implementation

felixboehm:
need PM input and challenge of requirements @pmaier1

PVince81:
PR here https://github.com/owncloud/customgroups/pull/82

PR #82, opened 2017-08-04
Prevent duplicate display names

PVince81:
Prevents duplicate display names

Fixes https://github.com/owncloud/customgroups/issues/77

- [x] backend prevention
- [x] option to allow duplicates
- [x] catch duplicate error when creating group
- [x] catch duplicate error when renaming group
    - [x] make field reappear, etc
    - [x] REQUIRES: PROPPATCH error detection https://github.com/owncloud/core/pull/28589

### DIFF
--- a/js/GroupsView.js
+++ b/js/GroupsView.js
@@ -42,7 +42,7 @@
 			}
 
 			this.collection.on('request', this._onRequest, this);
-			this.collection.on('sync destroy', this._onEndRequest, this);
+			this.collection.on('sync destroy error', this._onEndRequest, this);
 			this.collection.on('add', this._onAddModel, this);
 			this.collection.on('change', this._onChangeModel, this);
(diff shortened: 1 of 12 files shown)


sharidas:
Shared my observation:
- So once a group is created it does prevent creation of new group with same displayname. 
cg_dup_message

- Another thing I have noticed ( pardon me if its a wrong test I did ), Lets say if we create two customgroups```test one``` and ```test two```. Now rename ```test two``` to ```test one```. It does allow: 
cg_dup

PVince81:
@sharidas if you refresh the page you should see that the rename actually failed.

You need to check out this branch from core https://github.com/owncloud/core/pull/28589 so that the PROPPATCH error detection works correctly.

sharidas:
Aah I was thinking what am I doing wrong. Yes, refresh the page and no duplicates. :+1:

PR #84, opened 2017-08-17
Run tests against stable10

PVince81:


DeepDiver1975:
But not anymore against master?

Usually we add a second config - see https://github.com/owncloud/contacts/blob/25f745bde43ee6dcedb6860273078e3c363492ed/.travis.yml#L36-L39

PVince81:
@DeepDiver1975 hmm I didn't know that...

PVince81:
what is this ?

```
No composer.json in current directory, do you want to use the one at /home/travis/build/owncloud/core? [Y,n]? 
```

@DeepDiver1975 @SergioBertolinSG

DeepDiver1975:
the yaml file is not valid

````
 Remove log Raw log

ERROR: An error occured while trying to parse your .travis.yml file.

Please make sure that the file is valid YAML.

http://lint.travis-ci.org can check your .travis.yml.

The log message was: Build config file had a parse error: could not find expected ':' while scanning a simple key at line 22 column 9.

DeepDiver1975:
> what is this ?

we had the same on the guests app

PVince81:
there's a composer.json both in the app folder and in tests/integration, so not sure why it's not finding it

DeepDiver1975:
> there's a composer.json both in the app folder and in tests/integration, so not sure why it's not finding it

it tries to call composer in core's tests/integration - there was a composer file some time ago but it got merged into core's composer file

PVince81:
@DeepDiver1975 I'm totally confused now. Seems I've added a second matrix block.

Seems we already have a matrix block.

Are we running **all** tests against **both** stable10 and master or only the unit tests ?

At this point with integration tests I'd also expect to run them against both branches ?

DeepDiver1975:
> Are we running all tests against both stable10 and master or only the unit tests ?
> 
> At this point with integration tests I'd also expect to run them against both branches ?

this depends on the app and what we expect to be tested.
its a trade off between execution time and test everything ...

PVince81:
@DeepDiver1975 finally green, can you have a quick check that I did everything right ?

Issue #33, opened 2017-03-08
Better UI for switching permission


PVince81:
permission change should be a dropdown instead of pencil

PVince81:
I hope you can help here @felixheidecke

PVince81:
(next week, nothing urgent)

PVince81:
low prio, moving to "planned"

SergioBertolinSG:
![problems_roles_customgroups](https://user-images.githubusercontent.com/9048796/29564844-bd20b68a-8743-11e7-9506-4282afba78a8.gif)

PVince81:
ok, so it's "just" a bug with the tooltip

SergioBertolinSG:
And the ghost spinner at the right of the box for adding users.

PVince81:
I suggest we fix the tooltip separately because changing this to a dropdown as per the original post is much more work.

SergioBertolinSG:
There is another bug related with this UI. When changing the role of a group admin to member while being a group admin (having another available), the page should be refreshed, because there are elements which only a group admin should see and they require a refresh to dissapear.

PVince81:
For the tooltip and spinner issues, I fixed this in this new PR: https://github.com/owncloud/customgroups/pull/97

We should keep this ticket here open because the original request is to provide a dropdown instead lf a simple click.

PR #97, opened 2017-08-24
Improve spinners in members view

PVince81:
Spinner for autocomplete now appears when loading autocomplete list or
adding member.
Spinner for list appears when loading list only.
New spinner appears when toggling role.
Tooltip now properly removed when toggling role

For https://github.com/owncloud/customgroups/issues/33#issuecomment-324007120

### DIFF
--- a/css/app.css
+++ b/css/app.css
@@ -78,6 +78,14 @@
 	max-width: 1px;
 }
 
+#customgroups .loading-list {
+	margin-top: 10px;
+}
+
+#customgroups .icon-loading-small {
+	margin-left: 10px;
+}
+
 .customgroups-autocomplete-item {
 	display: flex;
 }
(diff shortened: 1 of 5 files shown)


SergioBertolinSG:
When removing yourself as admin of a group and changing your mind, the spinner spins forever.

spinning_forever

PVince81:
@SergioBertolinSG fixed, please try again

SergioBertolinSG:
Works.

A bit nitpicking: The profile pictures of the members of the groups blink when changing its role.

PVince81:
> A bit nitpicking: The profile pictures of the members of the groups blink when changing its role.

I saw this, likely no quick fix because it's related to how the avatar code works: basically it needs to do a server call to find out whether the user has an avatar. Would require some kind of caching.

This is something for another time.

PR #103, opened 2017-08-29
Add leave group dispatcher event

sharidas:
Add leave group dispatcher event.
Also trigger removeUserFromGroup event
when different users are deleted. That
is for users other than self.

### DIFF
--- a/lib/Dav/MembershipNode.php
+++ b/lib/Dav/MembershipNode.php
@@ -137,10 +137,15 @@ class MembershipNode implements \Sabre\DAV\INode, \Sabre\DAV\IProperties {
 		if ($currentUserId !== $userId) {
 			// only notify when the removal was done by another user
 			$this->helper->notifyUserRemoved($userId, $this->groupInfo, $this->memberInfo);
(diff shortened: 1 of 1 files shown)


lib/Dav/MembershipNode.php
```diff
@@ -137,10 +137,15 @@ public function delete() {
...
 			// only notify when the removal was done by another user
 			$this->helper->notifyUserRemoved($userId, $this->groupInfo, $this->memberInfo);
+			$event = new GenericEvent(null, ['user_displayName' => $userId, 'group_displayName' => $this->groupInfo['display_name']]);
+			$this->dispatcher->dispatch('\OCA\CustomGroups::removeUserFromGroup', $event);
 		}
 
-		$event = new GenericEvent(null, ['user_displayName' => $userId, 'group_displayName' => $this->groupInfo['display_name']]);
-		$this->dispatcher->dispatch('\OCA\CustomGroups::removeUserFromGroup', $event);
+		//Send dispatcher event if the removal is self
+		if ($currentUserId === $userId) {
```
sharidas: In this case self removal becomes leave group. Apart from self removal, other operations will be logged as delete from group. Let me know if this is ok?

Issue #90, opened 2017-08-22
LDAP users are not removed from dropdown when they are already members of the custom group.


SergioBertolinSG:
### Steps to reproduce

0. Set up a ldap server.
1. Create a custom group 'ldap_group' using a user from the ldap server.
2. Add ldap users to the custom group 'ldap_group'.
3. Try to add an already added ldap user to the custom group 'ldap_group'.

### Expected behaviour
Users which are already members of the customgroup doesn't appear in the dropdown.

### Actual behaviour
Ldap users which are already members of the customgroup still appear in the dropdown.
When choosing them, they cannot be added though, which is correct.

![screen shot 2017-08-22 at 16 49 54](https://user-images.githubusercontent.com/9048796/29571835-9bcd0cce-875a-11e7-8241-d312ca9bd0f8.png)

### Server configuration

**Operating system**:
Ubuntu 16.04

**Web server:**
Apache

**Database:**
MySQL

**PHP version:**
7.0

**ownCloud version:** (see ownCloud admin page)
10.0.3beta1
{"installed":"true","maintenance":"false","needsDbUpgrade":"false","version":"10.0.3.0","versionstring":"10.0.3 beta","edition":"Community","productname":"ownCloud"}

**Updated from an older ownCloud or fresh install:**
Fresh

**The content of config/config.php:**

```

```

**Are you using external storage, if yes which one:** local/smb/sftp/...
No.

**Are you using encryption:** 
No.

**Logs**
```
```

### Client configuration

**Browser**
Chrome

cc @PVince81

PVince81:
There is some custom code to exclude already existing users from the result list. Maybe there is a user id format mismatch there.

PVince81:
Only happen in customgroups autocomplete, not in the sharing one.

PVince81:
Argh, now I see that the code is completely wrong.

The search for existing members is only checking if there is a user_id that matches the pattern and is already a member of the custom group. This means it only ever works with user ids, which in LDAP are usually those unreadable UUIDs.

Steps to reproduce without LDAP:

1. Create a user "user1" with display name "One"
1. Create a custom group "customgroup1"
1. Add "One" to "customgroup1"
1. Try adding "One" again

Since "One" does not match the user_id "user1", the code does return that user as available choice.

There are two possible fixes:
1) Fix the backend to also join on the user account table to be able to resolve the display name
or
[...]

PR #106, opened 2017-09-11
Use IUserManager::find() when searching for members

PVince81:
This makes sure that it will not only search the user id or display name
but also any search terms that might be defined for LDAP users.

Fixes https://github.com/owncloud/customgroups/issues/90

### DIFF
--- a/lib/Service/MembershipHelper.php
+++ b/lib/Service/MembershipHelper.php
@@ -223,13 +223,10 @@ class MembershipHelper {
 	 * Return the user ids of the members in the given group matching the given pattern
 	 *
 	 * @param int $groupId numeric group id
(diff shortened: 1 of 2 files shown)


PVince81:
- [x] TEST: with DB user and display names: https://github.com/owncloud/customgroups/issues/90#issuecomment-328522325
- [x] TEST: with LDAP users where user id is not the same as display names: https://github.com/owncloud/customgroups/issues/90#issue-251985629

phil-davis:
@PVince81 a rebase of this should make Jenkins happy now. (ref #107 #108 )

SergioBertolinSG:
Works 👍

PVince81:
stalled build... will restart

PR #111, opened 2017-09-15
Bump version to 0.3.5

PVince81:


PVince81:
@phil-davis thank you, sharp eyes. I left the "Unreleased" link which is invisible anyway, it will be useful once we re-add that section.

phil-davis:
@DeepDiver1975 found one in my files_texteditor changelog an hour ago, so I was aware to look for it.

PR #115, opened 2017-10-19
Added a requirement coming from core

SergioBertolinSG:
To be able to use command line functions in core a variable was required in the config file. That has to be modified in all apps using the same code.

Coming from https://github.com/owncloud/core/pull/29265

### DIFF
--- a/tests/integration/config/behat.yml
+++ b/tests/integration/config/behat.yml
@@ -17,6 +17,7 @@ default:
               - admin
             regular_user_password: 123456
             mailhog_url: http://127.0.0.1:8025/api/v2/messages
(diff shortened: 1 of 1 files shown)


SergioBertolinSG:
OK the job failing is using core stable10 branch. Should we backport https://github.com/owncloud/core/pull/29265 to stable10 ?

phil-davis:
Yes, at the moment we need to keep stable10=master for all test infrastructure. Then app integration or UI tests see the same testing infrastructure when run against core master or stable10. 

One day, when we get to 10.1 11.0 etc core we will have to sort out app compatibility, and app test compatibility.

Issue #116, opened 2017-11-08
User adding autocomplete should comply with sharing restrictions


pmaier1:
When the global sharing restriction "Restrict users to only share with users in their groups" is active the "add users to a custom group" dialog autocomplete should only offer users that I'm allowed to share with (exactly like the sharing dialog autocomplete, but only for users, not groups or other share types).

PVince81:
Same for the enumeration checkboxes.

PVince81:
@pmaier1 fix is implemented

PR #117, opened 2017-11-13
Restrict autocomplete results when sharing restrictions in place

PVince81:
If sharee enumeration is disabled, only allow exact matches.
If sharee enumeration limited to groups, only search in user's groups.
If shares limited to group members, only search in user's groups.

Fixes https://github.com/owncloud/customgroups/issues/116

- [x] add unit tests


PVince81:
Not too happy about the solution, but it's how core does it as well...

PVince81:
- [ ] TODO: fix API call for adding members to refuse to add outside members when restriction is in place

PVince81:
Unit tests added

lib/Controller/PageController.php
```diff
@@ -80,7 +112,53 @@ public function searchUsers($group, $pattern, $limit = 200) {
...
+			$userGroups = $this->groupManager->getUserGroupIds($this->userSession->getUser());
+			foreach ($userGroups as $userGroup) {
+				$usersTmp = $this->groupManager->findUsersInGroup($userGroup, $pattern, $limit);
+				// merge and deduplicate results
+				foreach ($usersTmp as $uid => $user) {
+					$results[$uid] = $user;
+				}
+			}
+
+			// filter out existing members
```
jvillafanez: I'd rather move this filtering to another private function

lib/Controller/PageController.php
```diff
@@ -80,7 +112,53 @@ public function searchUsers($group, $pattern, $limit = 200) {
...
+				return !in_array($result->getUID(), $customGroupMembers);
+			});
+
+			$results = array_values($results);
+		} else {
+			$results = $this->helper->searchForNewMembers($groupInfo['group_id'], $pattern, $limit);
+		}
+
+		if (!$shareeEnumeration) {
+			// filter by exact match
```
jvillafanez: Same here, move to another function.

lib/Controller/PageController.php
```diff
@@ -80,7 +112,53 @@ public function searchUsers($group, $pattern, $limit = 200) {
...
+			$lowerSearch = strtolower($pattern);
+			$results = array_filter(
+				$results,
+				function($result) use ($lowerSearch) {
+					return
```
jvillafanez: How many results do we expect to handle in this `array_filter`? In terms of performance, this is probably the fastest solution, but the readability is bad. Unless we expect to handle thousands of results, I'd prefer to go for readability instead of performance because the gain would likely be unnoticeable.

lib/Service/MembershipHelper.php
```diff
@@ -360,6 +360,34 @@ public function canCreateGroups() {
 	}
 
 	/**
+	 * Checks whether the current user is allowed to add a member into the target group
+	 *
```
jvillafanez: Enumerating the cases will likely help to know what are the expectations in the future. I find weird that the super admin can bypass the `shareWithGroupOnly` restriction (or so it seems).

jvillafanez:
I'm a bit worried about the number of times we're traversing the result array in the `searchUser` function. I think we could traverse it 5 times in the worst case, and if we consider the array might be big, the performance drop might be noticeable.

lib/Service/MembershipHelper.php
PVince81: In general for custom groups the super admin is able to do whatever, even manage groups in which they aren't member. So being able to add any member sounds right.

But now I noticed that I'd need to also adjust the autocomplete to deliver all results.

@pmaier1 what are your thoughts on this ?

Note that an admin is usually also limited by the sharing restrictions in the regular autocomplete dialog. So maybe we should just use the same restriction here, still.

PVince81:
@jvillafanez so far we've set the limit to about 200 entries. I don't think it will be more. I expect that someone who uses the autocomplete to refine the search by typing more letters. In the search autocomplete in core I think we also do not have pagination, so you only get 20+ results or so.

I'll see if I can somehow reduce the number of iterations.

lib/Service/MembershipHelper.php
PVince81: I've removed this now

PVince81:
@jvillafanez I improved the impl, this should minimize the number of iterations already.

I'm sure we can likely improve this further if we had an `$exactMatch` param in IUserManager and IGroupManager find methods.

I also considered moving all these search related functions into a SearchService at some point.

Please let me know if this is suitable.

Some open TODOs:

- [ ] add limit unit test for the remaining cases

jvillafanez:
Looks good 👍

PVince81:
Added tests for pagination and squashed.

PVince81:
@jvillafanez quick look at the unit tests ? search for "Limit" in PageControllerTest.

I also slightly changed the limit handling in searchByMembershipGroup to make it continue searching until we have enough entries up to the limit, or there are no more entries found.

tests/unit/PageControllerTest.php
```diff
@@ -41,44 +45,324 @@ class PageControllerTest extends \Test\TestCase {
...
+		$user3 = $this->makeUser('user3', 'User Three');
+
+		$this->config->method('getAppValue')
+			->will($this->returnValueMap([
+				['core', 'shareapi_only_share_with_group_members', 'no', 'no'],
+				['core', 'shareapi_allow_share_dialog_user_enumeration', 'yes', 'yes'],
+				['core', 'shareapi_share_dialog_user_enumeration_group_members', 'no', 'no'],
+			]));
+
+		$this->handler->expects($this->at(0))
```
jvillafanez: I'd rather avoid the `$this->at(..)` if possible.

tests/unit/PageControllerTest.php
```diff
@@ -41,44 +45,324 @@ class PageControllerTest extends \Test\TestCase {
...
+		$response = $this->pageController->searchUsers('group1', 'us', 20);
+		$data = $response->getData();
+
+		$this->assertTrue(isset($data['results']));
+		$this->assertCount(20, $data['results']);
+		$this->assertEquals(
+			['userId' => 'user1', 'displayName' => 'User 1'],
+			$data['results'][0]
+		);
+		$this->assertEquals(
```
jvillafanez: This seems like a partial check. Maybe we can generate all the expected results and check all.

tests/unit/PageControllerTest.php
```diff
@@ -41,44 +45,324 @@ class PageControllerTest extends \Test\TestCase {
...
+			['userId' => 'user21', 'displayName' => 'User 21'],
+			$data['results'][19]
+		);
+	}
+
+	public function exactSearchDataProvider() {
+		$singleResult = [
+			['userId' => 'user1', 'displayName' => 'User One'],
```
jvillafanez: Should we include more data, such as email and search terms?

jvillafanez: hmmmm, it's set up in the test... not sure what's the best option here.

tests/unit/PageControllerTest.php
```diff
@@ -90,4 +374,222 @@ public function testSearchUsers() {
...
+					['group2', 'us', 20, 0, array_chunk($allUsers, 10)[1]],
+					// third group empty
+					['group3', 'us', 20, 0, ['user21' => $allUsers['user21']]],
+				],
+			],
+		];
+	}
+
+	/**
+	 * @dataProvider groupDataProvider
```
jvillafanez: This doesn't seem to be used

tests/unit/PageControllerTest.php
PVince81: will replace with `once()`, seems to be some remains from a previous construct

tests/unit/PageControllerTest.php
PVince81: Ok, I found a quick way to generate results and remove user3 from it

tests/unit/PageControllerTest.php
PVince81: This is the result data. I don't think we need to add extra info here. I'm not even sure the original code will contain extra data here. We do have a core PR where we add additional info for autocomplete results. We don't have anything similar yet for custom groups.

tests/unit/PageControllerTest.php
PVince81: scroll up, the function above is `groupDataProvider`

PVince81:
@jvillafanez thanks for the review. I've adjusted the parts I commented on and squashed.

tests/unit/PageControllerTest.php
jvillafanez: I checked the `$allUsers` variable below, which isn't used in the function, but the `$groupData` is used. No need to change anything here.

PR #138, opened 2018-03-08
Refactor to separate app code logic from core groups/memb logic + keep groups always in sync for centr_groups feature

mrow4a:
The goal of following changes were to:

Refactor:
- App code is only to serve user interface, and all interaction with core is through custom group manager (CustomGroupManager class) and GroupInterface (group backend). No DAV (user interface related) component should ever call CORE directly. 

Always sync with centr_groups:
- App keeps its groups/memberbers in sync with core with any action in the UX (add group/user remove group/user) 
 - Central groups in core will result in separate core groups/membership logic so that core never hits the app code (and thus core is more stable)

- [x] Manual checks passed with stable10 in core
- [x] Manual checks passed with centr_groups in core
- [x] Adjusted unit tests

### DIFF
--- a/appinfo/info.xml
+++ b/appinfo/info.xml
@@ -37,7 +37,7 @@ Sharing with a Custom Group is as easy and quick as always. Just click on the sh
 	</types>
 	<use-migrations>true</use-migrations>
 	<dependencies>
-		<owncloud min-version="10.0.3" max-version="10.1" />
+		<owncloud min-version="10.1" max-version="10.1" />
 	</dependencies>
 	<sabre>
 		<plugins>
--- a/lib/Controller/PageController.php
+++ b/lib/Controller/PageController.php
@@ -23,6 +23,7 @@ namespace OCA\CustomGroups\Controller;
 
 use OCP\AppFramework\Controller;
 use OCP\AppFramework\Http\TemplateResponse;
+use OCP\IL10N;
 use OCP\IUser;
 use OCP\AppFramework\Http;
 use OCP\AppFramework\Http\DataResponse;
@@ -60,6 +61,9 @@ class PageController extends Controller {
 	 */
 	private $userManager;
 
+	/** @var IL10N */
+	private $l10n;
+
 	public function __construct(
 		$appName,
 		IRequest $request,
@@ -67,7 +71,8 @@ class PageController extends Controller {
 		IUserSession $userSession,
 		IUserManager $userManager,
 		IGroupManager $groupManager,
-		CustomGroupsDatabaseHandler $handler
+		CustomGroupsDatabaseHandler $handler,
+		IL10N $l10n
 	) {
 		parent::__construct($appName, $request);
 		$this->handler = $handler;
@@ -75,6 +80,7 @@ class PageController extends Controller {
 		$this->userSession = $userSession;
 		$this->userManager = $userManager;
 		$this->groupManager = $groupManager;
+		$this->l10n = $l10n;
 	}
 
 	/**
@@ -91,6 +97,8 @@ class PageController extends Controller {
 
 	/**
 	 * Search in all groups the current user is member of
+	 *
+	 * TODO: Refactor to use custom groups database instead and reduce complexity
 	 *
 	 * @param string $customGroupId custom group id
 	 * @param string $pattern lowercase pattern
@@ -241,7 +249,7 @@ class PageController extends Controller {
 
 		$pattern = strtolower($pattern);
 
-		$groupInfo = $this->handler->getGroupByUri($group);
+		$groupInfo = $this->handler->getGroupBy('uri', $group);
 		if (is_null($groupInfo)) {
 			return new DataResponse(
 				[
--- a/lib/CustomGroupsBackend.php
+++ b/lib/CustomGroupsBackend.php
@@ -53,7 +53,7 @@ class CustomGroupsBackend implements \OCP\GroupInterface {
 	 * @return boolean
 	 */
 	public function implementsActions($actions) {
-		return ($actions & (self::GROUP_DETAILS | self::DELETE_GROUP)) !== 0;
+		return ($actions & (self::GROUP_DETAILS)) !== 0;
 	}
 
 	/**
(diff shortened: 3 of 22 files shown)


lib/Controller/PageController.php
```diff
@@ -241,8 +236,9 @@ public function searchUsers($group, $pattern, $limit = 200) {
 
 		$pattern = strtolower($pattern);
 
-		$groupInfo = $this->handler->getGroupByUri($group);
+		$groupInfo = $this->groupManager->get($group);
 		if (is_null($groupInfo)) {
+			// TODO makes no sense, l10n does not exist here..
```
mrow4a: Should I add l10n?

lib/Controller/PageController.php
```diff
@@ -92,19 +85,21 @@ public function index() {
 	/**
 	 * Search in all groups the current user is member of
 	 *
-	 * @param string $customGroupId custom group id
+	 * TODO: Get explanation what this was supposed to do
```
mrow4a: I have no idea what is this function about

lib/CustomGroupsBackend.php
```diff
@@ -47,42 +46,13 @@ public function __construct(
 	}
 
 	/**
-	 * Checks if backend implements actions.
+	 * Expose to the core only group details
```
mrow4a: This backend exposes to core only groupDetails

lib/CustomGroupsBackend.php
```diff
@@ -47,42 +46,13 @@ public function __construct(
...
-	}
-
-	/**
-	 * Checks whether the user is member of a group or not.
-	 *
-	 * @param string $uid uid of the user
-	 * @param string $gid gid of the group
-	 * @return boolean true if user is in group, false otherwise
-	 */
-	public function inGroup($uid, $gid) {
```
mrow4a: all memberships in this backend are gone to core - only groups are kept for consistency.

mrow4a:
Hey @IljaN @PVince81 could I have comments on the justification of change and review of the code? I updated the description

Thanks!

lib/Dav/GroupMembershipCollection.php
```diff
@@ -103,16 +103,9 @@ public function __construct(
...
-			throw new Forbidden("No permission to delete group \"$groupId\"");
-		}
+		$uri = $this->groupInfo['uri'];
 
-		$group = $this->groupManager->get('customgroup_' . $this->groupInfo['uri']);
-		if ($group === null) {
-			throw new NotFound("Group not found \"$groupId\"");
-		}
-		$group->delete();
+		$this->helper->deleteGroup($uri);
```
mrow4a: Moved the delete group logic to Helper (which does the sync to core also)

lib/Dav/GroupMembershipCollection.php
```diff
@@ -183,34 +176,9 @@ public function getProperties($properties) {
...
-
-		if (!$this->groupsHandler->addToGroup($userId, $groupId, CustomGroupsDatabaseHandler::ROLE_MEMBER)) {
-			throw new PreconditionFailed("The user \"$userId\" is already member of this group");
-		}
-
-		$this->helper->notifyUser($userId, $this->groupInfo);
-
-		$event = new GenericEvent(null, ['groupName' => $this->groupInfo['display_name'], 'user' => $userId]);
-		$this->dispatcher->dispatch('\OCA\CustomGroups::addUserToGroup', $event);
+		$this->helper->addUser($this->groupInfo['uri'], $userId);
```
mrow4a: Moved the add user to group logic to Helper (which does the sync to core also)

lib/Dav/GroupMembershipCollection.php
```diff
@@ -290,23 +258,8 @@ public function childExists($userId) {
 	 * @return boolean|int true or status code
 	 */
 	public function updateDisplayName($displayName) {
-		if (!$this->helper->isUserAdmin($this->groupInfo['group_id'])) {
-			return 403;
-		}
+		$result = $this->helper->updateGroup($this->groupInfo['uri'], $displayName);
```
mrow4a: Moved the group update logic to Helper (which does the sync to core also)

lib/Dav/GroupsCollection.php
```diff
@@ -140,17 +136,7 @@ public function createExtendedCollection($name, Mkcol $mkCol) {
...
-		}
-
-		$groupId = $this->groupsHandler->createGroup($name, $displayName);
-		if (is_null($groupId)) {
-			throw new MethodNotAllowed("Group with uri \"$name\" already exists");
-		}
-
-		// add current user as admin
-		$this->groupsHandler->addToGroup($this->helper->getUserId(), $groupId, true);
+		$this->helper->createGroup($name, $displayName);
```
mrow4a: Moved the group create logic to Helper (which does the sync to core also)

mrow4a:
Keep note that I also keep custom groups and memberships table - this is not to affect stability of the app with big rewrite (with core mem/groups being new), and additionaly for consistency so if something goes really wrong, some cronjob sync will adjust it back.

mrow4a:
@PVince81 Good idea! Will do :> 

Yes. it will work only with central_groups branch. However, without this change custom groups will also work, but users will see their memberships and new groups on login or when sync cron runs.

mrow4a:
@PVince81 @IljaN @tomneedham @butonic Just tested this app both with unit testss (>90% coverage) and manualy for stable10 core with 1st commit. This refactor should make this app cleaner, @IljaN thanks for earlier comments, they were very helpful - now there is a class CustomGroupsManager. 

Will work next for integration with centr_groups and adjusting unit tests.

mrow4a:
Yes, after change in first commit, this app will work with both central groups and with earlier versions. @PVince81

appinfo/info.xml
```diff
@@ -37,7 +37,7 @@ Sharing with a Custom Group is as easy and quick as always. Just click on the sh
 	</types>
 	<use-migrations>true</use-migrations>
 	<dependencies>
-		<owncloud min-version="10.0.3" max-version="10.1" />
+		<owncloud min-version="10.1" max-version="10.1" />
```
mrow4a: @PVince81 ok, with second commit I needed to increase the requirement, since createGroupFromBackend is needed for syncing with core. Everything else works as previously.

mrow4a: First commit however passed all checks with stable core

mrow4a:
central group tables wont be done

PR #141, opened 2018-03-29
Remove /ocs/ from end of acceptance tests baseUrl

phil-davis:
To work with core changes:
https://github.com/owncloud/core/pull/30943


phil-davis:
Note: this needs the change in ``core`` then run CI again.

PR #154, opened 2018-08-07
Use ocPath to testing app

phil-davis:
- Use ocPath to testing app
- remember and set default capabilities, to keep core methods happy
- update scenario step texts of core steps that were changed recently

### DIFF
--- a/tests/acceptance/features/bootstrap/CustomGroupsContext.php
+++ b/tests/acceptance/features/bootstrap/CustomGroupsContext.php
@@ -356,7 +356,13 @@ class CustomGroupsContext implements Context, SnippetAcceptingContext {
 	/**
 	 * Abstract method implemented from Core's FeatureContext
 	 */
-	protected function resetAppConfigs() {}
+	protected function resetAppConfigs() {
+		// Remember the current capabilities
+		$this->getCapabilitiesCheckResponse();
+		$this->savedCapabilitiesXml[$this->getBaseUrl()] = $this->getCapabilitiesXml();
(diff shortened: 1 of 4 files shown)


tests/acceptance/features/bootstrap/CustomGroupsContext.php
```diff
@@ -356,7 +356,13 @@ public function theSabreHTTPStatusCodeAnsweredShouldBe($statusCode) {
 	/**
 	 * Abstract method implemented from Core's FeatureContext
 	 */
-	protected function resetAppConfigs() {}
+	protected function resetAppConfigs() {
+		// Remember the current capabilities
+		$this->getCapabilitiesCheckResponse();
+		$this->savedCapabilitiesXml[$this->getBaseUrl()] = $this->getCapabilitiesXml();
+		// Set the required starting values for testing
+		$this->setCapabilities($this->getCommonSharingConfigs());
```
phil-davis: I need to add some capabilities stuff so that core ``restoreParametersAfterScenario()`` has something in:
```
$this->savedCapabilitiesChanges[$this->getBaseUrl()]
```
otherwise it complains about undefined index.

So I let it set the common sharing configs, just for the heck of it.

phil-davis:
This gets the acceptance tests passing again.
After this PR is merged, I can try to remove the local ``run.sh`` in this app and use the one from ``core``
It is a bit more complex in this app, because it has makefile stuff, and Travis does ``make test-acceptance``

PR #156, opened 2018-08-09
Trigger CI

phil-davis:
this is expected to fail the API acceptance tests, because the core QA tarballs now have the skeleton dir in a different place.


phil-davis:
fails as expected

PR #158, opened 2018-08-14
Remove local composer and run.sh

phil-davis:
My initial objective was to stop using the local ``tests/acceptance/run.sh`` here, delete it, and delete the ``composer.json`` that has a bunch of dev things in it that are in core anyway.
It is complicated because this repo has a ``Makefile`` that defines each of the ``tests-acceptance`` etc commands. I tried to minimize touching that.
Also, the CI is still run on Travis. So that needed a little massaging. In particular, I ended up removing the code coverage report (which does not appear as a required item here anyway), because it needs more massaging of Travis to get the ``phar`` that does the code coverage upload.


phil-davis:
Where should I go with this?

This could be merged, then I can work on migrating it to drone, and in the process bring back the code coverage report. If I do that, then what should I do with all the ``Makefile`` infrastructure here - use it from drone (which will make the ``.drone.yml`` different from other apps), or delete it, or leave it behing?

Or I could close this PR and just make some absolutely minimal change to get rid of the local ``run.sh`` and use the core ``run.sh`` - and leave behind all the other bits that drive the unit tests...

individual-it:
I would suggest to merge this PR, move to drone, and get rid of acceptance-tests in the Makefile, we do not use it anywhere else

phil-davis:
The ``Makefile`` also has entries for unit tests... 
drone in core and other apps generally does not do stuff like ``make test-php`` or ``make test-acceptance`` - it does direct commands that run ``phpunit` or ``run.sh`` or whatever.

Are we aiming for consistent build/test script infrastructure across core and apps?

If so, what do we want?
e.g.
- ``Makefile`` with all the build and test things in it
- developers run tests by typing ``make test-xxx``
- CI run tests by doing ``make test-xxx`` commands
- ``Makefile`` is responsible for sorting out how to do it.

Or some version of what we have now?

Maybe there is or should be a separate issue for this question - it is not just ```customgroups``

patrickjahns:
> Maybe there is or should be a separate issue for this question - it is not just ```customgroups``

Please open in QA - will write down the thoughts and proposal there - and yes I agree that we want to have consistent behavior 

Some quick notes for moving forward in the future:

CLI tools for tests will be moved into "namespaces" with composer-bin to not pollute the application composer dependencies and allow the usage of different php versions. ( phpstan requires php7.1 - but can analyze code for various php versions ). 

Secondly - the test rules will be moved into `Makefile` rules again. Any CI should in the end just call the Makefile rule.

.travis.yml
```diff
@@ -47,14 +53,16 @@ matrix:
       env: DB=mysql;CORE_BRANCH=master
     - php: 7.2
       env: DB=mysql;CORE_BRANCH=master
-    # PHP unit tests + code coverage
+    # PHP unit tests with mySQL
+    - php: 7.1
+      env: DB=mysql;TEST=test-php;CORE_BRANCH=master
     - php: 5.6
-      env: DB=mysql;TEST=test-upload-coverage
```
DeepDiver1975: what happend to this?

.travis.yml
```diff
@@ -47,14 +53,16 @@ matrix:
...
-    # PHP unit tests
+      env: DB=mysql;TEST=test-php
+    # PHP unit tests with pgSQL
     - php: 7.1
       env: DB=pgsql;TEST=test-php;CORE_BRANCH=master
     - php: 5.6
-      env: DB=pgsql;TEST=test-php;CORE_BRANCH=stable10
```
DeepDiver1975: where did this one go?

phil-davis: This one just loses ``CORE_BRANCH=stable10`` because that is the default - it makes it look like the other matrix entries that do not mention a branch.

.travis.yml
phil-davis: It is gone - see my comments above about "I ended up removing the code coverage report"

Decision is needed about if this gets merged, then I proceed with changing CI to drone (and bring back coverage) or...

And then the whole discussion about what the "standard" app CI config should be anyway.

phil-davis:
Issue #159 raised to move to drone and put back code coverage.

PR #161, opened 2018-08-20
Remove use BasicStructure from local acceptance test context

phil-davis:
and use methods from core ``FeatureContext`` instead.

### DIFF
--- a/tests/acceptance/features/bootstrap/CustomGroupsContext.php
+++ b/tests/acceptance/features/bootstrap/CustomGroupsContext.php
@@ -133,7 +133,7 @@ class CustomGroupsContext implements Context, SnippetAcceptingContext {
 		$exists = false;
 		foreach ($customGroupsList as $customGroupPath => $customGroupName) {
 			if ((!empty($customGroupName))
-				&& (array_values($customGroupName)[0] == $customGroup)
+				&& (\array_values($customGroupName)[0] == $customGroup)
 			) {
 				$exists = true;
 			}
@@ -159,7 +159,7 @@ class CustomGroupsContext implements Context, SnippetAcceptingContext {
 		$exists = false;
 		foreach ($customGroupsList as $customGroupPath => $customGroupName) {
 			if ((!empty($customGroupName))
-				&& (array_values($customGroupName)[0] == $displayName)
+				&& (\array_values($customGroupName)[0] == $displayName)
 			) {
 				$exists = true;
 				break;
@@ -185,7 +185,7 @@ class CustomGroupsContext implements Context, SnippetAcceptingContext {
 		$exists = false;
 		foreach ($customGroupsList as $customGroupPath => $customGroupName) {
 			if ((!empty($customGroupName))
-				&& (array_values($customGroupName)[0] == $customGroup)
+				&& (\array_values($customGroupName)[0] == $customGroup)
 			) {
 				$exists = true;
 			}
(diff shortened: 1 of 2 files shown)


phil-davis:
Note: this will fail today, because ``getDavPath`` is private in ``FeatureContext``. That is fixed in a core PR and will appear in the core QA tarball overnight.

tests/acceptance/features/bootstrap/CustomGroupsContext.php
```diff
@@ -135,103 +179,150 @@ public function customGroupShouldNotExist($customGroup){
...
 		$client->setThrowExceptions(true);
 		$appPath = '/customgroups/groups/';
-		$fullUrl = $this->getBaseUrl() . '/' . $this->davPath . $appPath . $customGroup;
+		$fullUrl = $this->featureContext->getBaseUrl() . '/' . $this->featureContext->getDavPath() . $appPath . $customGroup;
 		try {
-			$response = $client->proppatch($fullUrl, $properties, 1);
-			$this->response = $response;
-			return $response;
-		} catch (\Sabre\HTTP\ClientException $e) {
-			$this->response = null;
```
individual-it: how do we set the response in `BasicStructure` now? don't we need `$this->featureContext->setResponse($response);`?

tests/acceptance/features/bootstrap/CustomGroupsContext.php
```diff
@@ -135,103 +179,150 @@ public function customGroupShouldNotExist($customGroup){
...
+	 * @return bool
+	 * @throws ClientException
+	 */
 	public function sendProppatchToCustomGroupMember($userRequesting, $customGroup, $userRequested, $properties = null){
-		$client = $this->getSabreClient($userRequesting);
+		$client = $this->featureContext->getSabreClient($userRequesting);
 		$client->setThrowExceptions(true);
 		$appPath = '/customgroups/groups/';
-		$fullUrl = $this->getBaseUrl() . '/' . $this->davPath . $appPath . $customGroup . '/' . $userRequested;
+		$fullUrl = $this->featureContext->getBaseUrl() . '/' . $this->featureContext->getDavPath() . $appPath . $customGroup . '/' . $userRequested;
```
individual-it: please make that line shorter

tests/acceptance/features/bootstrap/CustomGroupsContext.php
```diff
@@ -135,103 +179,150 @@ public function customGroupShouldNotExist($customGroup){
...
 		try {
-			$response = $client->proppatch($fullUrl, $properties, 1);
-			$this->response = $response;
```
individual-it: same as above, don't we need to change `BasicStructure` `$response`?

tests/acceptance/features/bootstrap/CustomGroupsContext.php
phil-davis: This is strange code which does a Sabre client thing. The responses from that are not the same thing as in FeatureContext. The previous code would sometimes make a mess of FeatureContext response, but it was just lucky that steps afterward did not try to use the FeatureContext method to check the HTTP response.

tests/acceptance/features/bootstrap/CustomGroupsContext.php
phil-davis: See comment above - the return value of proppatch is not the sort of response that FeatureContext has.

phil-davis:
@individual-it I pushed another commit that makes the code match the acceptance test code style "standard".

PR #162, opened 2018-08-21
Specify PHP 5.6 in composer platform requirements

phil-davis:
to ensure that composer comes up with dependencies that can work on PHP 5.6


phil-davis:
Somewhat scarily, ``composer`` comes up with Guzzle v6 to satisfy an ``ocular`` dependency, which is just used with Scrutinizer code coverage upload. It should not effect run-time app code.

Issue #159, opened 2018-08-14
Move CI to drone


phil-davis:
Move CI  for ``customgroups`` to drone.

Note: code coverage on Travis got temporarily knobbled in #158 - remember to bring it.

While implementing drone, it might be easy to also get ``php-cs-fixer`` going. See issue #145

phil-davis:
@dpakach I assigned you to this issue because you did PR #172 
I guess you are going to do more PRs to finish this issue?

PR #173, opened 2018-11-12
Add .drone.yml file and fix makefile

dpakach:
Add ```.drone.yml``` file with tests and fix the makefile to automatically install dependencies in running `make test-php-style`

Related to #159


.drone.yml
```diff
@@ -0,0 +1,328 @@
...
+    when:
+      matrix:
+        TEST_SUITE: api-acceptance
+
+  codecov:
+    image: plugins/codecov:2
+    secrets: [codecov_token]
+    pull: true
+    paths:
+      - tests/unit/clover.xml
```
DeepDiver1975: path holding a file?

composer.json
```diff
@@ -10,9 +10,15 @@
     },
     "require-dev": {
 		"phpunit/phpunit": "^5.7",
-		"scrutinizer/ocular": "^1.3.1"
+        "scrutinizer/ocular": "^1.3.1",
```
DeepDiver1975: ocular is not needed any more

PR #184, opened 2018-11-20
Remove .travis.yml file as tests are now in drone

### LINKED ISSUE #159: Move CI to drone

dpakach:
All CI tests have moved to drone now. So, .travis.yml can be deleted now.

Fixes #159


phil-davis:
@DeepDiver1975 this repo needs drone to be set to "required" and Travis set to "not required".

.drone.yml
```diff
@@ -165,6 +183,18 @@ matrix:
     - PHP_VERSION: 7.1
       TEST_SUITE: owncloud-coding-standard
 
+    - PHP_VERSION: 7.1
+      OC_VERSION: daily-master-qa
+      TEST_SUITE: codecheck
+      NEED_SERVER: true
```
phil-davis: Maybe codecheck can be done without installing the server - remove the lines NEED_SERVER and NEED_INSTALL_APP?

Otherwise you could try having another flag FIX_PERMISSIONS to control the fix-permissions pipeline step separately to NEED_SERVER. Maybe it will be able to install the server and app, not fix-permissions, and the code check will work.

dpakach: That test actually runs an `occ` command so I think we will need server but we can try adding that flag for fixing permission

Issue #175, opened 2018-11-14
Unit tests failing in oracle database


dpakach:
when running unit tests in oracle database in drone this error occurs
```
1) OCA\CustomGroups\Tests\unit\CustomGroupsDatabaseHandlerTest::testGetGroupsByDisplayName
Failed asserting that two strings are equal.
--- Expected
+++ Actual
@@ @@
-'my_group_1'
+'my_group_3'
```

The drone link for this error is [here](https://drone.owncloud.com/owncloud/customgroups/6/53)

PVince81:
hmmm, this sometimes happen if the unit tests are relying too much on database result order.
I might have missed doing this when writing them and it is known from experience that sometimes Oracle returns results out of order...

PVince81:
@dpakach I assume you discovered this as part of your work in https://github.com/owncloud/customgroups/pull/173 ?

and perhaps we did not enable Oracle before so the issue was not discovered ?

in the future please link relevant tickets :-)

dpakach:
@PVince81 yes, the tests which are in travis never used oracle so it might have missed before

PVince81:
Fix is here: https://github.com/owncloud/customgroups/pull/187

PR #187, opened 2018-11-26
Sort groups when requested from DB

PVince81:
This makes for deterministic output, especially on Oracle.

Fixes https://github.com/owncloud/customgroups/issues/175


PVince81:
@DeepDiver1975 @dpakach CI is green now, it was indeed the sort order in Oracle.

please approve

PR #197, opened 2019-01-30
Add validation for group creation

sharidas:
While creating custom groups few
validation added. The validation include:
- group names cannot begin with space
- group name cannot be empty string or null
- minimum length for group name is 2

### DIFF
--- a/js/GroupsView.js
+++ b/js/GroupsView.js
@@ -268,7 +268,12 @@
 						OC.Notification.showTemporary(t('customgroups', 'A group with this name already exists'));
 						return;
 					}
-					OC.Notification.showTemporary(t('customgroups', 'Could not create group'));
+					if (response.status === 422) {
+						OC.Notification.showTemporary(t('customgroups', "The group name can not be empty or start with space. The group name should at least have 2 characters"));
+					}
+					if (response.status === 403) {
+						OC.Notification.showTemporary(t('customgroups', 'Could not create group'));
+					}
 				}
 			});
 		},
--- a/lib/Dav/GroupsCollection.php
+++ b/lib/Dav/GroupsCollection.php
@@ -21,6 +21,7 @@
 
 namespace OCA\CustomGroups\Dav;
 
+use OCA\CustomGroups\Exception\ValidationException;
 use Sabre\DAV\IExtendedCollection;
 use Sabre\DAV\MkCol;
 use Sabre\DAV\Exception\NotFound;
@@ -119,6 +120,21 @@ class GroupsCollection implements IExtendedCollection {
 	public function createExtendedCollection($name, Mkcol $mkCol) {
 		if (!$this->helper->canCreateGroups()) {
 			throw new Forbidden('No permission to create groups');
+		}
+
+		/** Group name cannot be empty */
+		if (($name === '') || ($name === null)) {
+			throw new ValidationException('Can not create empty group');
+		}
+
+		/** Group name must be at least 2 character long */
+		if (\strlen($name) < 2) {
+			throw new ValidationException('The group name should be at least 2 characters long.');
+		}
+
+		/** Group name should not start with space */
+		if ($name[0] === ' ') {
+			throw new ValidationException('The group name can not start with space');
 		}
 
 		$displayName = $name;
--- a/lib/Exception/ValidationException.php
+++ b/lib/Exception/ValidationException.php
@@ -1,0 +1,36 @@
+<?php
+/**
+ * @author Sujith Haridasan <sharidasan@owncloud.com>
+ *
+ * @copyright Copyright (c) 2019, ownCloud GmbH
+ * @license AGPL-3.0
+ *
+ * This code is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License, version 3,
+ * as published by the Free Software Foundation.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU Affero General Public License for more details.
+ *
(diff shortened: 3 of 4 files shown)


js/GroupsView.js
```diff
@@ -268,6 +268,16 @@
...
 						return;
 					}
+					if (response.status === 403) {
+						var parser = new DOMParser();
+						var xmlDoc = parser.parseFromString(response.body, "text/xml");
+						if ((xmlDoc.getElementsByTagName("s:message")[0] !== undefined) &&
+							(xmlDoc.getElementsByTagName("s:message")[0].childNodes[0].nodeValue !==
+							'No permission to create groups')) {
+							OC.Notification.showTemporary(t('customgroups',
+								xmlDoc.getElementsByTagName("s:message")[0].childNodes[0].nodeValue));
```
jvillafanez: I don't think this works as intended...
* Comparing message string is a bad practice that can easily go wrong. We need to change the approach
* You're translating (`t(....)`) a variable. I don't think this will be picked up by the translation mechanism, and it can also go wrong because the message can be quite different that the ones translated.

Unless you want to show more information, I expect the 403 status means that you don't have permissions to create groups, so just check the status code and show a translatable "You aren't allowed to create groups" or similar. I expect no additional meaning for the 403 status.

lib/Dav/GroupsCollection.php
```diff
@@ -121,6 +125,22 @@ public function createExtendedCollection($name, Mkcol $mkCol) {
 			throw new Forbidden('No permission to create groups');
 		}
 
+		/** Group name cannot be empty */
+		if (($name === '') || ($name === null)) {
+			throw new Forbidden($this->l10n->t('Can not create empty group'));
```
jvillafanez: I'd advise against translating exception messages. There are currently several problems:
* Most of the exceptions (if not all) contains untranslated messages. Adding translated messages in the exceptions is inconsistent, and it also will be confusing in the future: should I translate the message or not? when should I translate?
* You need to inject the l10n class inside classes that aren't part of the frontend and could be far or deep. Why would a class that won't interact directly with any client needs this dependency to work? Translations are better handled in controllers or related classes.

jvillafanez: In addition, any better alternative than a Forbidden exception? I'm not sure if it's suited for a validation failure.

lib/Dav/GroupsCollection.php
```diff
@@ -121,6 +125,22 @@ public function createExtendedCollection($name, Mkcol $mkCol) {
...
+		}
+
+		/** Group name must be at least 2 character long */
+		if (\strlen($name) < 2) {
+			throw new Forbidden($this->l10n->t('The group name should be at least 2 characters long.'));
+		}
+
+		/** Group name should not start with space */
+		\preg_match('/^(\s+).*/', $name, $matches);
```
jvillafanez: just check the first char of the string. Something like `$name[0] !== ' '` performs better and is easier to understand.

js/GroupsView.js
PVince81: shouldn't that rather be 400 bad request ? it's not about missing privileges to create such group but really that the request is invalid.

another possible code is 422 unprocessable entity when meeting validation errors

in this one case I think I'd rather tend to 400 bad request

lib/Dav/GroupsCollection.php
PVince81: the way we did this was usually to just return the status code and let the client generate the message

in some cases the client can also parse the exception class from the XML body to find out about details

I suggest that here we also add client-side verification and don't rely on the server.
Server-side verification is then only here for people using the API directly.

js/GroupsView.js
sharidas: To be honest, I did not liked this idea too. But it was my mistake, I should have asked this in the ticket. Damn I missed to do it :( Thanks for the helping hand :)

jvillafanez: I'd vote for a 422 for validation errors

js/GroupsView.js
```diff
@@ -268,7 +268,13 @@
 						OC.Notification.showTemporary(t('customgroups', 'A group with this name already exists'));
 						return;
 					}
-					OC.Notification.showTemporary(t('customgroups', 'Could not create group'));
+					if (response.status === 422) {
+						OC.Notification.showTemporary(t('customgroups', "The group name can not be emtpy or start with space. "+
```
jvillafanez: typo -> `empty`
I'm not sure if the translation mechanism will properly pick the string concatenation, I'd rather stay on the safe side and use only one string.

lib/Dav/GroupsCollection.php
```diff
@@ -71,6 +72,9 @@ class GroupsCollection implements IExtendedCollection {
 	*/
 	private $dispatcher;
 
+	/** @var \OCP\IL10N  */
+	private $l10n;
```
jvillafanez: This isn't needed

js/GroupsView.js
sharidas: Replaced string concatenation and made as single string.

lib/Dav/GroupsCollection.php
sharidas: Removed the  `$l10n`.

sharidas:
> One question that comes to my mind: there might be more validation to be done in the future. How are we planning to extend this? I see 2 options:
> 
> * Change the `InvalidGroupNameException` to `ValidationException` to make more generic in order to reuse the exception in other future places
> * Create more new exceptions which might share the same http code.
> 
> I'm not sure what is the best option in this case

I would vote for the first one. A more generic one.  The http status code returned would be still `422`. Any deviation from the validation exception would help us write new type of exception which ould return a different http status code.

davitol:
screen shot 2019-03-06 at 12 38 08
@sharidas It seems it is possible to rename custom  group names and bypass the validations

PR #206, opened 2019-02-28
Make acceptance tests run and pass

phil-davis:
The API acceptance tests were not actually running in drone because the pipeline step had a section with `comands` spelt wrong, so the step did nothing and did not fail.

1) fix spelling of `commands:`
2) add `WebDavPropertiesContext` to `behat.yml` - this is needed because of a core refactoring in recent months
3) Add a `getSabreClient` method to `CustomGroupsContext`. That method was removed in core. Put it locally here so we can get the tests passing again, then we can refactor it away, like what was done in core.

### DIFF
--- a/.drone.yml
+++ b/.drone.yml
@@ -110,7 +110,7 @@ pipeline:
       - TEST_SERVER_URL=http://owncloud
       - PLATFORM=Linux
       - BEHAT_SUITE=${BEHAT_SUITE}
-    comands:
+    commands:
       - make test-acceptance-api
     when:
       matrix:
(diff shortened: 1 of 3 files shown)


individual-it:
`PHP Fatal error: Call to undefined method FeatureContext::getSabreClient() in /var/www/owncloud/apps/customgroups/tests/acceptance/features/bootstrap/CustomGroupsContext.php on line 110`

phil-davis:
Yes, we refactored out the Sabre client stuff from core. I need to find that and do something here in `customgroups` tests.

PR #209, opened 2019-03-11
Add validation for group rename

sharidas:
Added validation for group rename.

### DIFF
--- a/lib/Dav/GroupMembershipCollection.php
+++ b/lib/Dav/GroupMembershipCollection.php
@@ -22,6 +22,7 @@
 namespace OCA\CustomGroups\Dav;
 
 use OCA\CustomGroups\CustomGroupsDatabaseHandler;
+use OCA\CustomGroups\Exception\ValidationException;
 use Sabre\DAV\PropPatch;
 use Sabre\DAV\Exception\MethodNotAllowed;
 use Sabre\DAV\Exception\NotFound;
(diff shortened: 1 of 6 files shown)


sharidas:
Status of this PR:
- [x] Verified rename option available in the UI. Found only one option 
test1
- Updated the text for 422 response code in the UI. That is the oc notification is updated for the rename of groups.
- [x] Updated the unit tests

davitol:
Screen Shot 2019-03-11 at 10 35 16


Validation now works on renaming groups, but user gets twice the yellow warning

davitol:
Also it keeps been possible to create groups named with a single character using special ones like ñ or ç

Screen Shot 2019-03-11 at 10 38 38

sharidas:
> Also it keeps been possible to create groups named with a single character using special ones like ñ or ç
Ok multibyte ones... Will update the PR.

davitol:
multibyte charactes bug is fixed, but it still can be posible to create a group name with a dot.

Screen Shot 2019-03-12 at 09 57 15

sharidas:
> multibyte charactes bug is fixed, but it still can be posible to create a group name with a dot.
> 
> Screen Shot 2019-03-12 at 09 57 15

The reason for this is hidden in this piece of code https://github.com/owncloud/customgroups/blob/master/js/GroupsView.js#L259-L263. This is the reason why `.` ( dot ) is accepted. While debugging i could see the index is incremented and hence the associated index is appended with the group name.

sharidas:
Updated the PR, which will catch the dot `.` mentioned https://github.com/owncloud/customgroups/pull/209#issuecomment-471912828

davitol:
> Updated the PR, which will catch the dot `.` mentioned [#209 (comment)](https://github.com/owncloud/customgroups/pull/209#issuecomment-471912828)

👍 Tested this update too and works

lib/Dav/GroupMembershipCollection.php
```diff
@@ -292,9 +293,28 @@ public function childExists($userId) {
...
+			throw new ValidationException('Can not rename to empty group');
+		}
+
+		/**
+		 * Verify if the character length is less than 2 or if its multibyte string, then
+		 * verify if the multibyte character length is less than 2.
+		 */
+		if ((\strlen($displayName) < 2) ||
+			((\mb_strlen($displayName, 'UTF-8') !== \strlen($displayName)) &&
+				\mb_strlen($displayName, 'UTF-8') < 2)) {
```
PVince81: according to https://coderwall.com/p/qgeuna/php-string-length-the-right-way using `mb_strlen` should be enough.

I also wonder about many other places in the core code where we don't check this properly... (tech debt ticket in core?)

sharidas: Replaced `strlen` with `mb_strlen`.

PVince81: Can you validate all test cases that @davitol mentionned ? or have them covered through unit tests (if not already) ?

PVince81: with "validate" I mean "confirm that it is tested and works"

sharidas:
With the current state of the PR below are the tests done:
- [x]  Group name with single dot `.` , is not allowed to create. 
singledot

- [x]  Group name with single multibyte character not allowed to create 
singlemb

- [x] With 2 unicode characters its possible to create custom group 
2charmb

- [x] After creating a custom group renaming with single char not possible
     - Remove the last character from previous screen shot 
Rename_to_singlemb
     - Cannot rename the group name `ÄÜ` to single dot.
     - Cannot rename the group name `ÄÜ` to group name `1`
     - Cannot rename the group name `ÄÜ` to `a`

This covers the issues raised in this PR by @davitol . I do have unit tests added for the issue raised in this PR.
I confirm this pr is tested and works as expected.

PR #245, opened 2019-09-02
Initial drone starlark

phil-davis:
This is the same `starlark` code as other apps that have been converted.
It has specific config for `customgroups` CI.

### DIFF
--- a/.drone.starlark
+++ b/.drone.starlark
@@ -1,0 +1,971 @@
+config = {
+	'app': 'customgroups',
+	'rocketchat': 'builds',
+
+	'branches': [
+		'master'
+	],
+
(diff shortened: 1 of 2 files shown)


.drone.starlark
```diff
@@ -0,0 +1,971 @@
...
+			]
+		},
+		'reducedDatabases' : {
+			'phpVersions': [
+				'7.1',
+				'7.2',
+				'7.3',
+			],
+			'databases': [
+				'sqlite',
```
individual-it:
```suggestion
				'pgsql',
```

phil-davis:
@individual-it ready for review again

PR #271, opened 2019-11-20
App should only be shown if it is whitelisted for guest

sharidas:
Customgroups app should be shown for guest only
if the app is whitelisted by the guest app.

### DIFF
--- a/lib/SettingsPanel.php
+++ b/lib/SettingsPanel.php
@@ -21,15 +21,30 @@
 
 namespace OCA\CustomGroups;
 
+use OCP\IConfig;
+use OCP\IGroupManager;
+use OCP\IUser;
+use OCP\IUserSession;
 use OCP\Settings\ISettings;
 use OCP\Template;
 use OCA\CustomGroups\Service\MembershipHelper;
 
 class SettingsPanel implements ISettings {
+	/** @var MembershipHelper */
 	private $helper;
+	/** @var IGroupManager */
+	private $groupManager;
+	/** @var IUserSession */
+	private $userSession;
+	/** @var IConfig */
+	private $config;
 
-	public function __construct(MembershipHelper $helper) {
+	public function __construct(MembershipHelper $helper, IGroupManager $groupManager,
+								IUserSession $userSession, IConfig $config) {
 		$this->helper = $helper;
+		$this->groupManager = $groupManager;
+		$this->userSession = $userSession;
+		$this->config = $config;
 	}
 
 	public function getPanel() {
@@ -46,6 +61,27 @@ class SettingsPanel implements ISettings {
 	}
 
 	public function getSectionID() {
+		/**
+		 * Check if this app should be shown or not if the user belongs to the disallowed
+		 * groups in system config. If the user belongs to the disallowed groups
+		 * in system config, then lets not show this app in the personal settings
+		 * page of the user.
+		 */
+		$user = $this->userSession->getUser();
+		$disallowedGroups = $this->config->getSystemValue('customgroups.disallowed-groups', null);
+		if ($user !== null && $disallowedGroups !== null) {
+			foreach ($disallowedGroups as $disallowedGroup) {
+				$group = $this->groupManager->get($disallowedGroup);
+				if ($group === null) {
+					continue;
+				}
+
+				if ($this->groupManager->isInGroup($user->getUID(), $group->getGID())) {
+					return '';
+				}
+			}
+		}
+
 		return 'customgroups';
 	}
 }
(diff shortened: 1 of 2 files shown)


sharidas:
#### Description
The guest user when navigates to the settings page, is able to see customgroups app even though the customgroups app is not whitelisted for the guest. This change does bring inter-app dependency with the patch, because this app now has to see if the guest app settings are available or not. For instance in this PR, I have wrote the code to check if the guest group exists or not and if it exists check if the logged in user is part of the guest user. And finally if the conditions are true, then it checks if customgroups app is whitelisted by the admin for the guest user. If it is whitelisted, then show the customgroups app, else don't.

#### Motivation
The customgroups app should not be shown in the guest user settings page, if the app is not whitelisted for the guest user.

#### Issue
- enterprise ticket 3526

#### Test done
- Enable the guest app
- As a default guest user navigate to the settings page, the customgroups app will not be visible.
- Now add customgroups to whitelist. And the app should be visible to the guest user settings page.

jvillafanez:
I don't really like this approach. There are several important drawbacks:
* This breaks the "no inter-app dependencies" rule. The customgroups app needs to have access to where the guests app has stored the whitelist. The problem is that guest could change the location, and this will break the app again (nobody should care about where this is stored, only the guests app, which is the responsible)
* I expect all the apps to have the same behaviour. While this is happening for guests, this could happen with any app (files_external, files_classifier, etc). This means that will have to patch in a similar way all the apps.
* Following the previous point, this PR only patches the settings page. While this could be fine for customgroups, other apps might need additional changes (external endpoints, for example).
* This PR only hides the settings page, but the code is still available for the guests. This means that, even though the guests can't use the UI (in the settings page) to create custom groups, they can still create custom groups by replicating the requests. Depending on how the customgroups app works, this could be unusable, but still it's something to take into account for other apps.

lib/SettingsPanel.php
```diff
@@ -46,6 +66,33 @@ public function getPriority() {
...
 
 	public function getSectionID() {
+		/**
+		 * Check if this app should be shown or not if the user is guest
+		 * If the logged in user is a guest user and the guest group exist
+		 * then customgroups app should only be shown if the app is whitelisted
+		 * for guest user. Else it shouldn't be shown.
+		 */
+		$user = $this->userSession->getUser();
+		$group = $this->config->getAppValue('guests', 'group', 'guest_app');
```
DeepDiver1975: this is not what I was talking about - let me explain again:

1.) define a new system config customgroups.disallowed-groups
2.) hide the system panel if the current user is in one of these disallowed groups
3.) done

micbar: @DeepDiver1975 The customers expect the whitelist mechanism to work.
How can we make this work.

micbar: It should work without additional config.

sharidas: @DeepDiver1975 I have a query here 
> 1.) define a new system config customgroups.disallowed-groups

Just for my understanding, where does customgroups fetch the info of the disallowed groups?
I am writing down some of the cases which we shouldn't miss here:
- the guest app can change the group name.
- the group whitelist members can also change.

In this scenario I am wondering how to handle the new systemconfig?

jvillafanez: all the apps would need to implement something similar, so we'd need `customgroups.disallowed-groups`, `files_external.disallowed-groups`, `oauth.disallowed-groups`, etc, and all these lists must be manually maintained by the admin, and all of them could be different than the one that the guests app is (or should be) providing.

micbar: @jvillafanez @sharidas 
We will not invest more time to fix the whitelist.
Please go forward with the config switch. `customgroups.disallowed-groups`

micbar: I know this is not the best solution, but we should not invest more time here. 
@DeepDiver1975 FYI

sharidas: Updated the PR with `customgroups.disallowed-groups`. @micbar @jvillafanez

lib/SettingsPanel.php
```diff
@@ -46,6 +61,27 @@ public function getPriority() {
...
+		/**
+		 * Check if this app should be shown or not if the user belongs to the disallowed
+		 * groups in system config. If the user belongs to the disallowed groups
+		 * in system config, then lets not show this app in the personal settings
+		 * page of the user.
+		 */
+		$user = $this->userSession->getUser();
+		$disallowedGroups = $this->config->getSystemValue('customgroups.disallowed-groups', null);
+		if ($user !== null && $disallowedGroups !== null) {
+			foreach (\explode(',', $disallowedGroups) as $disallowedGroup) {
```
DeepDiver1975: why explode? system config can take an array right away. no need for any parsing

sharidas: Done. I thought of using comma, separated strings. But array make sense. Thanks.

PR #273, opened 2019-11-26
Deny admin access when system config is set

sharidas:
Deny admin access when system config is set
by the administrator. This will make admin
user behave as regular user for the custom groups.

### DIFF
--- a/lib/Dav/GroupsCollection.php
+++ b/lib/Dav/GroupsCollection.php
@@ -22,6 +22,7 @@
 namespace OCA\CustomGroups\Dav;
 
 use OCA\CustomGroups\Exception\ValidationException;
+use OCP\IConfig;
 use Sabre\DAV\IExtendedCollection;
 use Sabre\DAV\MkCol;
 use Sabre\DAV\Exception\NotFound;
@@ -60,6 +61,9 @@ class GroupsCollection implements IExtendedCollection {
 	 */
 	private $helper;
 
+	/** @var IConfig */
+	private $config;
+
 	/**
 	 * User id for which to use memberships or null for all groups
 	 *
@@ -73,20 +77,25 @@ class GroupsCollection implements IExtendedCollection {
 	private $dispatcher;
 
 	/**
-	 * Constructor
+	 * GroupsCollection constructor.
 	 *
+	 * @param IGroupManager $groupManager
 	 * @param CustomGroupsDatabaseHandler $groupsHandler custom groups handler
-	 * @param MembershipHelper $helper helper
+	 * @param MembershipHelper $helper
+	 * @param IConfig $config
+	 * @param string|null $userId
 	 */
 	public function __construct(
 		IGroupManager $groupManager,
 		CustomGroupsDatabaseHandler $groupsHandler,
 		MembershipHelper $helper,
+		IConfig $config,
 		$userId = null
 	) {
 		$this->groupManager = $groupManager;
 		$this->groupsHandler = $groupsHandler;
 		$this->helper = $helper;
+		$this->config = $config;
 		$this->userId = $userId;
 
 		$this->dispatcher = \OC::$server->getEventDispatcher();
@@ -215,7 +224,12 @@ class GroupsCollection implements IExtendedCollection {
 		if ($this->userId !== null) {
 			$allGroups = $this->groupsHandler->getUserMemberships($this->userId, $search);
 		} else {
-			$allGroups = $this->groupsHandler->getGroups($search);
+			$disallowAdminAccessAll = $this->config->getSystemValue('customgroups.disallow-admin-access-all', false);
+			if ($disallowAdminAccessAll) {
+				$allGroups = $this->groupsHandler->getUserMemberships($this->helper->getUserId(), $search);
+			} else {
+				$allGroups = $this->groupsHandler->getGroups($search);
+			}
 		}
 		return \array_map(function ($groupInfo) {
 			return $this->createMembershipsCollection($groupInfo);
--- a/lib/Service/MembershipHelper.php
+++ b/lib/Service/MembershipHelper.php
@@ -170,11 +170,26 @@ class MembershipHelper {
 	 * @return boolean true if the user can administrate, false otherwise
 	 */
 	public function isUserAdmin($groupId) {
-		// ownCloud admin is always admin of any custom group
-		if ($this->isUserSuperAdmin()) {
+		$isCurrentUserAdmin = $this->isUserSuperAdmin();
+		$memberInfo = $this->getUserMemberInfo($groupId);
+		if ($isCurrentUserAdmin) {
+			$denyAdminAccessAll = $this->config->getSystemValue('customgroups.disallow-admin-access-all', false);
+			/**
+			 * If system config customgroups.disallow-admin-access-all is set to true
+			 * then ownCloud admin is denied from the groups which it is not the
+			 * owner of or member of. It's also denied from the display name and editing members
+			 * of the group which its not not allowed to.
+			 */
+			if ($denyAdminAccessAll) {
+				return ($memberInfo !== null && $memberInfo['role']);
+			}
+			/**
+			 * If customgroups.disallow-admin-access-all is not set then ownCloud admin
+			 * has access to all (which is default behaviour)
+			 */
 			return true;
 		}
-		$memberInfo = $this->getUserMemberInfo($groupId);
+
 		return ($memberInfo !== null && $memberInfo['role']);
 	}
 
--- a/lib/Dav/RootCollection.php
+++ b/lib/Dav/RootCollection.php
@@ -22,6 +22,7 @@
 namespace OCA\CustomGroups\Dav;
 
 use OCA\CustomGroups\CustomGroupsDatabaseHandler;
+use OCP\IConfig;
 use Sabre\DAV\Exception\NotFound;
 use Sabre\DAV\Exception\MethodNotAllowed;
 use Sabre\DAV\SimpleCollection;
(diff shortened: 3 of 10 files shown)


sharidas:
#### Description
The idea here is to deny admin access the custom groups. Inorder to acheive this in this PR I have added a system config `customgroups.disallow-admin-edit`. The sys admin can provide the list of groups whose access is denied for the admin. This would disable admin of oC to deny modification of the group name and access of members from the group -> if the group name is added to the config key `customgroups.disallow-admin-edit`.

#### Issue
- Enterprise ticket 3501

#### Testing done
- When a custom group name is added to the system config `customgroups.disallow-admin-edit`
   - The admin cannot modify the group(s) which is added in `customgroups.disallow-admin-edit`
   - The admin cannot get the members of the group(s) in `customgroups.disallow-admin-edit`

lib/Service/MembershipHelper.php
```diff
@@ -170,9 +170,22 @@ private function getUserMemberInfo($groupId) {
 	 * @return boolean true if the user can administrate, false otherwise
 	 */
 	public function isUserAdmin($groupId) {
-		// ownCloud admin is always admin of any custom group
-		if ($this->isUserSuperAdmin()) {
-			return true;
+		$disallowedGroupsForAdmin = $this->config->getSystemValue('customgroups.disallow-admin-edit', null);
```
jvillafanez: double-check we want to store the information there. It might be fine if we want the "ops guy" or whoever is managing the host to manage that.

lib/Service/MembershipHelper.php
```diff
@@ -170,9 +170,22 @@ private function getUserMemberInfo($groupId) {
...
+		$isUserAdmin = $this->isUserSuperAdmin();
+		$groupInfo = $this->groupsHandler->getGroup((string)$groupId);
+		foreach ($disallowedGroupsForAdmin as $disallowedGroupForAdmin) {
+			if ($isUserAdmin && $groupInfo !== null && $disallowedGroupForAdmin === $groupInfo['display_name']) {
```
jvillafanez: Maybe we can change this a bit
```
if ($isUserAdmin) {
  foreach ($disallowedGroupsForAdmin......) {
    if ($groupInfo !== null && $disallowedGroupForAdmin === $groupInfo['display_name']) {
      return false;
    }
  }
}
```
I think this is easier to read.
Note that the `$this->config->getSystemValue('customgroups.disallow-admin-edit', null);` can also be moved inside the `if ($isUserAdmin)` (ifi the user isn't admin, no need to look for the groups), same for the `$groupInfo`

lib/Service/MembershipHelper.php
```diff
@@ -170,9 +170,22 @@ private function getUserMemberInfo($groupId) {
...
+				return false;
+			}
+		}
+
+		/**
+		 * If ownCloud admin is not denied from the groups listed in system config
+		 * customgroups.disallow-admin-edit, then:
+		 * ownCloud admin is always admin of any custom group
```
jvillafanez: good comment here :+1:

lib/Service/MembershipHelper.php
jvillafanez: my point here is, who is going to maintain this option? because new groups can be created and for some of them we might not want the admin to edit them.

sharidas: Its a valid point. But lets say if there is an option to automate this, then also the admin can see this in the config.php file, right?

lib/Dav/GroupsCollection.php
```diff
@@ -215,7 +221,9 @@ public function search(Search $search = null) {
 		if ($this->userId !== null) {
 			$allGroups = $this->groupsHandler->getUserMemberships($this->userId, $search);
 		} else {
-			$allGroups = $this->groupsHandler->getGroups($search);
+			$disallowAdminAccessAll = $this->config->getSystemValue('customgroups.disallow-admin-access-all', null);
```
jvillafanez: Following other options, I think it's better to make `false` as default and check for `true`. Note that `customgroups.disallow-admin-access-all => false` will be consider the same as setting it to true, which will cause problems.

It seems easier to read something like:
```
if ($disallowAdminAccessAll) {
  $allGroups = $this->groupHandler->......
} else {
  $allGroups = $this->groupHandler->....
}
```
In theory, the option could contain values that could be evaluated to true, but I don't think it's critical, so it's enough with the change above.

lib/Service/MembershipHelper.php
```diff
@@ -170,10 +170,19 @@ private function getUserMemberInfo($groupId) {
 	 * @return boolean true if the user can administrate, false otherwise
 	 */
 	public function isUserAdmin($groupId) {
```
jvillafanez: can we change the method name? the PHPDoc clarifies what the method does, but if I see `isUserAdmin('admin')` it isn't intuitive for me.

lib/Service/MembershipHelper.php
```diff
@@ -170,10 +170,19 @@ private function getUserMemberInfo($groupId) {
...
-		// ownCloud admin is always admin of any custom group
-		if ($this->isUserSuperAdmin()) {
-			return true;
+		$isUserAdmin = $this->isUserSuperAdmin();
```
jvillafanez: this is confusing with the method name. We have a `isUserAdmin` meaning that the current user is an admin, and a method `isUserAdmin` meaning that the current user can administer the target group.

lib/Service/MembershipHelper.php
```diff
@@ -170,10 +170,19 @@ private function getUserMemberInfo($groupId) {
...
+		$denyAdminAccessAll = $this->config->getSystemValue('customgroups.disallow-admin-access-all', null);
```
jvillafanez: Same as above, default to `false` and check for `true`. And move it inside the condition if the variable won't be used outside

jvillafanez:
While reviewing the PR I've noticed a thing that should be refactored:

Changes in the Users and Root collections (in this PR) shows that the config object is just being passing around. Neither the users nor the root collections are doing anything with the config object, and it just being used because they create a group collection at some point (which is where the config object is required). This should be refactored to use a factory and move the object creation there.

I don't think we have time to fix it here (it doesn't seem an easy change), but we should include it as a technical debt to be checked the sooner the better.

lib/Service/MembershipHelper.php
sharidas: I don't mind changing the method name, but there are instances where it would affect. For example the grep in the customgroups app shows:
```
lib/Dav/GroupMembershipCollection.php:          if (!$this->helper->isUserAdmin($groupId)) {
lib/Dav/GroupMembershipCollection.php:          if (!$this->helper->isUserAdmin($groupId)) {
lib/Dav/GroupMembershipCollection.php:          if (!$this->helper->isUserMember($groupId) && !$this->helper->isUserAdmin($groupId)) {
lib/Dav/GroupMembershipCollection.php:                  && !$this->helper->isUserAdmin($groupId)) {
lib/Dav/GroupMembershipCollection.php:          if (!$this->helper->isUserMember($groupId) && !$this->helper->isUserAdmin($groupId)) {
lib/Dav/GroupMembershipCollection.php:          if (!$this->helper->isUserAdmin($this->groupInfo['group_id'])) {
lib/Dav/MembershipNode.php:             if (!$this->helper->isUserAdmin($groupId)
lib/Dav/MembershipNode.php:             if (!$this->helper->isUserAdmin($groupId)) {
```

Would have to change in these locations too.

lib/Service/MembershipHelper.php
sharidas: Done

lib/Service/MembershipHelper.php
sharidas: Renamed the variable to `isCurrentUserAdmin`.

lib/Dav/GroupsCollection.php
sharidas: Done.

lib/Service/MembershipHelper.php
jvillafanez: pinging @micbar to decide. We can leave the method rename for a follow up PR, but we should make the PR soon.

lib/Service/MembershipHelper.php
```diff
@@ -170,10 +170,20 @@ private function getUserMemberInfo($groupId) {
...
+			$denyAdminAccessAll = $this->config->getSystemValue('customgroups.disallow-admin-access-all', false);
+			/**
+			 * If system config customgroups.disallow-admin-access-all is set to true
+			 * then ownCloud admin is denied from the groups which it is not the
+			 * owner of or member of. It's also denied from the display name and editing members
+			 * of the group which its not not allowed to.
+			 * If customgroups.disallow-admin-access-all is not set then ownCloud admin
+			 * has access to all (which is default behaviour)
+			 */
+			return !($denyAdminAccessAll === true);
```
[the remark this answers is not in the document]
sharidas: Slightly modified the condition here. Now the admin is able to add user(s) to the group which its owner of.

jvillafanez:
Nothing to add from me. If this is tested and works properly, I think we can merge it.

sharidas:
Another round of testing done with this PR:

Without the config setting:
- admin was able to add user(s) to the `user1Group`.
- Create a folder in `user1` and share to `user1Group`. This folder only accessible to users `user1` and `user2` ( as these are the members of the `user1Group`)

With the config setting:
- admin cannot see the `user1Group` ( as its not the member of the group )
- `user1` adds `admin` to `user1Group` and now admin can see this group. Now the `admin` can see the shared folder created by `user1`.
- Using curl try to update the display name of `user1Group` to `user5CreateGroup123`. Forbidden response is shown.
- Remove the admin from `user1Group`.
    - Try to remove any member of `user1Group` other than `user1` using curl as `admin` user. You should see message that its not possible. A Forbidden exception is thrown.
    - Try to add a new member to `user1Group` using curl as `admin` user. You should see message that its not possible. A Forbidden exception is thrown.

With the config option set, the admin does behave like normal or regular user. It can only access the groups which it has access to (Not ALL).

sharidas:
@micbar I have performed tests ( https://github.com/owncloud/customgroups/pull/273#issuecomment-558577172 and https://github.com/owncloud/customgroups/pull/273#issuecomment-559514978 )
- From UI and
- Also using curl. The reason for testing with curl is I wanted to make sure the system is not vulnerable when admin tries to attack the server.

Issue #306, opened 2020-02-06
Customgroups handlebars are conflicting with the ones in core


HanaGemela:
customgroups apps brings its own handlebars which are conflicting with the ones in core
customgroups > core
0.5.1 is compatible with 10.4

Expected result: customgroups handlebars https://github.com/owncloud/customgroups/blob/master/package.json#L14 match server version of handlebars

davitol:
* customgroups 0.5.1 installation from scratch + oC 10.3.2 => FAILS
* customgroups 0.5.1 upgrade from 0.4.X + oC 10.3.2 => FAILS

![Screen Shot 2020-02-06 at 10 54 11](https://user-images.githubusercontent.com/10596560/73926556-1e7b0e80-48d0-11ea-9bf7-97c1e1119f86.png)



** customgroups 0.5.1 installation from scratch + oC 10.4.0  RC1 => WORKS
** customgroups 0.5.1 upgrade from 0.5.0 + oC 10.4.0  RC1 => WORKS

phil-davis:
That is surprising to me (that it works with 10.4.0RC1), because 10.4.0RC1 does not have the latest/matching handlebars version.

I guess 10.4.0RC1 has a "good enough" handlebars version.

davitol:
> I guess 10.4.0RC1 has a "good enough" handlebars version.

Yes, that is also my thought. This PR https://github.com/owncloud/core/pull/36439  update handlebars for RC1 and that was good enough I guess

PR #308, opened 2020-02-06
Update owncloud min-version

HanaGemela:
Fixes https://github.com/owncloud/customgroups/issues/306


HanaGemela:
@phil-davis could you have a look at the failing CI?

phil-davis:
@micbar I changed your commit so that in the `config[]` section it now selects different settings for the servers to run acceptance tests against.

The actual starlark code from `def main()` down is the same in all apps (almost - user_ldap has some bonus code). I really want to keep it like that, and drive everything by data in the `config[]` section. That makes it much easier when we find a bug in the starlark code - we can fix it and then just paste to all other apps without thinking too much.

micbar:
@phil-davis Should we keep the tests runnning against daily master?

phil-davis:
> @phil-davis Should we keep the tests runnning against daily master?

yes, we need to run acceptance tests with some core version

micbar:
I meant we ran it against daily master and latest stable in the past, now we removed one platform. We could re enable it after the 10.4.0 release.

Issue #309, opened 2020-02-07
Run the tests against latest stable again 


HanaGemela:
@phil-davis please run the tests against latest stable again https://github.com/owncloud/customgroups/pull/308#issuecomment-583109286

phil-davis:
This will be blocked until 10.4 release, when the "latest" download link is for a 10.4 iamge.

phil-davis:
Unassigned myself - I don't want this in my assigned issues list until it gets to be unblocked

PR #316, opened 2020-03-04
Revert "test against daily server only"

phil-davis:
This reverts commit 81d85f3642f326efa545f7d2c2717324b5352211.

Fixes issue #309 

core `latest` is not 10.4.0

Note: removing these lines from `.drone.star` means that drone will now default to running the acceptance tests against both master and latest core.

### DIFF
--- a/.drone.star
+++ b/.drone.star
@@ -45,9 +45,6 @@ config = {
 			'suites': [
 				'apiCustomGroups'
 			],
-			'servers': [
-				'daily-master-qa'
-			],
 		},
 	}
 }


phil-davis:
drone https://drone.owncloud.com/owncloud/customgroups/1018/16/4
```
$ wget -qO- https://download.owncloud.org/community/daily/owncloud-latest.tar.bz2 | tar -xj -C /tmp/owncloud/ --strip 1
Fetched 10.4.0 - channel: stable - build at 2020-02-27T14:31:06+00:00 - SHA: 8b878ef9b40ba850b6ac3bf20250f2923262c0d5
```

It runs against `latest` which is `10.4.0` - good.

PR #370, opened 2020-09-12
[Tests-Only] Fixup 'last share id' test step

phil-davis:
The format of this test step was changed in core.


dpakach (approved):
LGTM :+1:

Issue #446, opened 2021-06-25
Review automated tests for sharing with custom groups


phil-davis:
There are already various test scenarios for creating shares to custom groups:
https://github.com/owncloud/customgroups/blob/master/tests/acceptance/features/apiCustomGroups/sharingCustomGroups.feature

But most of those scenarios just create the custom group and check that the status code in the response is good. They do not check that the share actually works for the share receiver.

- enhance the scenarios so that they check that the share receiver can use the share (download, upload, list, create a folder, delete)
- also check that permissions are obeyed (e.g. cannot upload or create folder in a read-only share)
- add scenarios for interactions with "normal" user and group shares (to make sure that the same resource can be shared to mixtures of users, groups and custom groups)

And think about other combinations of ways that custom groups can be used, and add reasonable test coverage.

(Can custom group shares have an expiration date? If so, then test it. If not, then test what happens when someone tries to set an expiration date)

SwikritiT:
@phil-davis all the PR's for this issue have been merged, I think we can close this.

phil-davis:
Agree - closing.

PR #447, opened 2021-06-30
[tests-only ] Add tests for sharing to check sharee can perform various operation on received share

SwikritiT:
part of https://github.com/owncloud/customgroups/issues/446

Changes made in this PR:
- remove test step that weren't being used in the scenario
- Add new feature file with tests to check that sharee can perform various operation in shared resources like
  - upload
  - download
  - create
  - delete
  - rename
  - move

### DIFF
--- a/tests/acceptance/features/apiCustomGroups/sharingCustomGroups.feature
+++ b/tests/acceptance/features/apiCustomGroups/sharingCustomGroups.feature
@@ -140,10 +140,8 @@ Feature: Sharing Custom Groups
       | /myFOLDER/myTMP/ |
 
   Scenario: Sharing again an own file while belonging to a custom group
-    Given as user "admin"
-    And user "Alice" has been created with default attributes and small skeleton files
+    Given user "Alice" has been created with default attributes and small skeleton files
     And user "Alice" has created a custom group called "sharing-group"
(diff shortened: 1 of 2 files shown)


tests/acceptance/features/apiCustomGroups/sharingCustomGroups.feature
```diff
@@ -140,10 +140,8 @@ Feature: Sharing Custom Groups
       | /myFOLDER/myTMP/ |
 
   Scenario: Sharing again an own file while belonging to a custom group
-    Given as user "admin"
-    And user "Alice" has been created with default attributes and small skeleton files
+    Given user "Alice" has been created with default attributes and small skeleton files
```
Talank, line 143: It might be better to create users without skeleton files.

SwikritiT: I think that whole file needs refactoring for skeleton files. I'll do it in another PR.

kiranparajuli589 (commented):
LGTM rather than existing comment

dpakach (commented):
lgtm 👍🏼

phil-davis (approved):

PR #521, opened 2022-06-21
feat: add guest invitation

DeepDiver1975:
Screenshot from 2022-06-21 10-27-12


pmaier1:
Please make sure that the CSV import will also work with guests (and with multiple users in general).

DeepDiver1975:
> Please make the input validation consistent with the regular sharing dialog. E.g., "ab@cd" offers guest creation while you need to add ".xy" in the sharing dialog.

We could now start a discussion what is the technical correct implementation....

... anyway will adjust the new code base for the sake of consistency

pmaier1:
> We could now start a discussion what is the technical correct implementation....

We've had that one earlier and decided for usability against technical correctness :)

JammingBen (approved):

js/MembersView.js
```diff
@@ -202,9 +202,16 @@
 			var $loading = this.$('.member-input-view .loading');
 			$loading.removeClass('hidden');
 
+			var displayName = data.displayName || userId
+			var role = null
+			if (data.type === 'guest') {
+				displayName = userId
```
JammingBen, line 208: [Comment from JP](https://github.com/owncloud/customgroups/pull/522#discussion_r919721406):

> we might need a comment here. The displayname should be already initialized above prioritizing the displayname over the userid. If guests will use the userid anyway I think this needs a comment to explain it's intentional, otherwise it might be a bug because one type of users will use the displayname while others the userid.

lib/Controller/PageController.php
```diff
@@ -280,10 +289,22 @@ public function searchUsers($group, $pattern, $limit = 200) {
 		$results = \array_map(function (IUser $entry) {
 			return [
 				'userId' => $entry->getUID(),
-				'displayName' => $entry->getDisplayName()
+				'displayName' => $entry->getDisplayName(),
+				'type' => 'user'
 			];
 		}, $results);
 
+		# guest integration
```
JammingBen, line 297: [Comment from JP](https://github.com/owncloud/customgroups/pull/522#discussion_r919730018):

I'm a bit confused here and maybe we need a comment clarifying.
Aren't guests shown as part of the regular result? or is this additional result a placeholder so the user can click on it to add that new user as a guest?

lib/Service/GuestIntegrationHelper.php
```diff
@@ -0,0 +1,155 @@
...
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License, version 3,
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>
+ *
+ */
+
+namespace OCA\CustomGroups\Service;
+
+use OCA\Guests\Controller\UsersController;
```
JammingBen, line 24: [Comment from JP](https://github.com/owncloud/customgroups/pull/522#discussion_r919737936):

> I'm not sure if this works without the guests app around...
I think it's better to use either an API call or even a symfony event in order to remove the dependency with the app.

JammingBen (commented):
I just noticed that JP added a few comments in https://github.com/owncloud/customgroups/pull/522 that relate to this PR. I added them here.

lib/Service/GuestIntegrationHelper.php
jnweiger: Guests is part of the *-complete.tar bundle, so we are 'pretty sure (tm)' it is installed.
API call (with proper error handling) is of course the better solution.

jnweiger:
Confirmed implemented in 0.7.0-rc.1

There is no problem when the guests app is disabled. The functionality is not available then. But no error messages seen due to `use OCA\Guests\Controller\UsersController;` as long as guests is installed, evertyhing is fine.

PR #522, opened 2022-06-22
Implement batch action for inviting users to groups

JammingBen:
Works towards https://github.com/owncloud/enterprise/issues/2865

### DIFF
--- a/js/MembersInputView.js
+++ b/js/MembersInputView.js
@@ -250,7 +250,7 @@
 						},
 					}).done(function (result) {
 						for (var j = 0; j < result.results.length; j++) {
-							var userToAdd = result.results[j]
+							var userToAdd = result.results[j];
 							var addUser = true;
 
 							// only add users that match exact with the search term


JammingBen:
This one is ready for review, sorry for not marking earlier. It needs #521 to be finished and merged first though, otherwise we permanently have merge conflicts with each force push to #521

phil-davis:
I put this into the Classic Server Team Board project, and mentioned some people for review.

1 JS test failed - I have no idea if that is a real fail.

There is a conflict reported, so that n eeds to be sorted out.

js/MembersInputView.js
```diff
@@ -149,11 +179,120 @@
...
+				)
+			}
+
+			return Promise.all(promises).then(function() {
+				for (var i = 0; i < users.length; i++) {
+					if (!users[i]) continue;
+					var user = users[i].trim();
+					var userAdded = false;
+
+					for (var j = 0; j < foundUsers.length; j++) {
```
jvillafanez, line 282: same as above, it seems better to use a set instead of an array

js/MembersView.js
```diff
@@ -202,9 +202,16 @@
 			var $loading = this.$('.member-input-view .loading');
 			$loading.removeClass('hidden');
 
+			var displayName = data.displayName || userId
+			var role = null
+			if (data.type === 'guest') {
+				displayName = userId
```
jvillafanez, line 208: we might need a comment here. The displayname should be already initialized above prioritizing the displayname over the userid. If guests will use the userid anyway I think this needs a comment to explain it's intentional, otherwise it might be a bug because one type of users will use the displayname while others the userid.

lib/Controller/PageController.php
```diff
@@ -280,10 +289,22 @@ public function searchUsers($group, $pattern, $limit = 200) {
...
 			return [
 				'userId' => $entry->getUID(),
-				'displayName' => $entry->getDisplayName()
+				'displayName' => $entry->getDisplayName(),
+				'type' => 'user'
 			];
 		}, $results);
 
+		# guest integration
+		if (!$results && $this->guestIntegrationHelper->canBeGuest($pattern)) {
```
jvillafanez, line 298: I'm a bit confused here and maybe we need a comment clarifying.
Aren't guests shown as part of the regular result? or is this additional result a placeholder so the user can click on it to add that new user as a guest?

lib/Service/GuestIntegrationHelper.php
```diff
@@ -0,0 +1,155 @@
...
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License, version 3,
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>
+ *
+ */
+
+namespace OCA\CustomGroups\Service;
+
+use OCA\Guests\Controller\UsersController;
```
jvillafanez, line 24: I'm not sure if this works without the guests app around...
I think it's better to use either an API call or even a symfony event in order to remove the dependency with the app.

js/MembersInputView.js
JammingBen: Things got mixed up here as this code belongs to https://github.com/owncloud/customgroups/pull/521. I transferred this comment to https://github.com/owncloud/customgroups/pull/521#pullrequestreview-1041452290.

js/MembersView.js
JammingBen: Things got mixed up here as this code belongs to https://github.com/owncloud/customgroups/pull/521. I transferred this comment to https://github.com/owncloud/customgroups/pull/521#pullrequestreview-1041452290.

lib/Service/GuestIntegrationHelper.php
JammingBen: Things got mixed up here as this code belongs to https://github.com/owncloud/customgroups/pull/521. I transferred this comment to https://github.com/owncloud/customgroups/pull/521#pullrequestreview-1041452290.

jvillafanez (approved):

jnweiger:
Confirmed implemented in 0.7.0-rc.1
image

Issue #535, opened 2022-09-08
[QA] Buttons for managing custom groups do not aling to browser size


mmattel:
The three buttons on top right should align properly dependent on frame window size.

It only shows correct, three buttons in a row, when on full screen...

The current layout is fixed and breakes the button `Import as CSV`. This is really bad UI.

Guessing that no code change is needed, but some CSS adoption.

<img width="703" alt="user_settings_custom_groups" src="https://user-images.githubusercontent.com/3321281/189078747-ccc31c1c-5da0-4f8c-8669-a9e2d7abe238.png">

@JammingBen @DeepDiver1975 maybe you can have a look on
@jnweiger @GeraldLeikam @pmaier1 fyi

jnweiger:
UI glitch, not a release blocker IMO

jnweiger:
In customgroups-0.7.1-rc.1 I can do this:

![image](https://user-images.githubusercontent.com/1108546/190814164-2d812235-78b2-4e62-85cf-34906ef4c00c.png)

* The issues is not fixed. the button still wraps,
* I can produce a scrollbar floating in mid air and overlaying text, when the height is too little.

Not sure if fixing all these things is worth the effort.

jnweiger:
Confirmed fixed in customgroups-0.7.1-rc.1

But then a scrollbar appeared floating in mid air, overlayed on the buttons, when the vertical space became to little....

![image](https://user-images.githubusercontent.com/1108546/190826410-d95c4c97-e655-46c9-90ef-5110f563237f.png)

mmattel:
At least the horizontal "standard" issue breaking the button is gone which was eye catching ugly.

More improvements are welcomed, but agreed that this should be done only when the effords are not high.

PR #536, opened 2022-09-16
Fix 'Import as CSV'-button alignment on small screens

JammingBen:
Fixes https://github.com/owncloud/customgroups/issues/535

image

### DIFF
--- a/css/app.css
+++ b/css/app.css
@@ -113,6 +113,7 @@
 
 .custom-group-import-label {
 	cursor: pointer;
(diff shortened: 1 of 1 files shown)


mmattel:
Are there any browser restrictions?

JammingBen:
> Are there any browser restrictions?

Not that I'm aware of. It's just a `display: inline-style`, that's supported by each browser.

mmattel (approved):

PR #564, opened 2023-04-13
fix: use ChromeHeadless instead of PhantomJS

DeepDiver1975:


DeepDiver1975:
js test passed -> merge

PR #610, opened 2023-11-14
fix: drop JavaScript function `tipsy()` in favor of botstrap tooltip

DeepDiver1975:
refs https://github.com/owncloud/core/pull/41099

can be released any time - but before 10.14 goes out ....

### DIFF
--- a/js/admin.js
+++ b/js/admin.js
@@ -7,6 +7,4 @@ $(document).ready(function() {
 		}
 		OC.AppConfig.setValue('customgroups', $(this).attr('name'), value);
 	});
-
-	$('.section .icon-info').tipsy({gravity: 'w'});
 });


phil-davis (approved):

js/admin.js
```diff
@@ -7,6 +7,4 @@ $(document).ready(function() {
 		}
 		OC.AppConfig.setValue('customgroups', $(this).attr('name'), value);
 	});
-
-	$('.section .icon-info').tipsy({gravity: 'w'});
```
jvillafanez, line 11: I'd expect a `$('.....').tooltip()` or something similar given the description. I assume this is intended.

jvillafanez (approved):

js/admin.js
DeepDiver1975: `$('.section .icon-info').tooltip()`  is already called in core - it was a duplicate form the first day

PR #611, opened 2023-11-14
Fix missing where clause in inGroupByUri()

IljaN:
User was reported to be in every custom-group as soon as he was in one custom-group by the inGroupByUri method.

http://sqlfiddle.com/#!17/9097f7/1

Incorrect resulting query

```sql
SELECT user_id 
FROM oc_custom_group_member m, oc_custom_group g 
WHERE uri = 'GroupFG' AND user_id = 'userE';
```

Correct/Intended query
```sql
SELECT user_id 
FROM oc_custom_group_member m, oc_custom_group g 
WHERE g.group_id = m.group_id AND uri = 'GroupFG' AND user_id = 'userE';
```

Slightly-Offtopic: This should probably be refactored in to a LEFT JOIN?

### DIFF
--- a/lib/CustomGroupsDatabaseHandler.php
+++ b/lib/CustomGroupsDatabaseHandler.php
@@ -95,7 +95,7 @@ class CustomGroupsDatabaseHandler {
 			->from('custom_group_member', 'm')
 			->from('custom_group', 'g')
 			->where($qb->expr()->eq('g.group_id', 'm.group_id'))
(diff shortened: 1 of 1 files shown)


jvillafanez (approved):

jvillafanez:
I agree we should check the performance of other variations. I guess the current one is fine because we don't expect a massive usage, but the cross join is usually very expensive.

DeepDiver1975:
Why was no test added? @IljaN

PR #626, opened 2024-04-09
fix: properly cleanup users upon removal

pako81:
We are currently missing the hook for removing the user from the `oc_custom_group_member` table when performing the user delete. I do not see any reason why the user should still be listed in the custom groups view (with an X avatar).

This would solve the issue for both "normal" and guest users, as currently when navigating through the group members views a new invitation mail is going to be resent to the guest user so that he/she will regain access to resources he/she should no longer have access to.

### DIFF
--- a/appinfo/app.php
+++ b/appinfo/app.php
@@ -19,9 +19,12 @@
  *
  */
 
+use OCA\CustomGroups\Hooks;
+
 $app = new \OCA\CustomGroups\Application();
 $app->registerGroupBackend();
 $app->registerNotifier();
+$app->getContainer()->query(Hooks::class)->register();
 
 if (!\defined('PHPUNIT') && !\OC::$CLI) {
 	$pathInfo = \OC::$server->getRequest()->getPathInfo();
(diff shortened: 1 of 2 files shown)


jnweiger (approved):
Looks good to me. 
My reproducer is fixed. Also I no longer see any ghost user with avatar (X).

I don't understand why CI fails, the case of removing a user from a custom group works fine for me, with or without this patch.

pako81:
I also do not understand it, but running the tests locally confirm that the test in failing with this PR in place.

appinfo/app.php
```diff
@@ -19,9 +19,12 @@
  *
  */
 
+use OCA\CustomGroups\Hooks;
+
 $app = new \OCA\CustomGroups\Application();
 $app->registerGroupBackend();
 $app->registerNotifier();
+$app->getContainer()->query(Hooks::class)->register();
```
IljaN, line 27: You should be able to call this like this: 

```suggestion
Hooks::register();
```

IljaN (approved):
See my suggestion, other than that LGTM.

jnweiger (approved):
Hooks:register(); is nicer.
Works like a charm.

PR #696, opened 2026-06-17
test: bump behat acceptance test dependencies

phil-davis:
Change the PHP version for running Behat to 8.3
Update the version of Behat and its dependencies that is used. Convert the Behat config from YML to a PHP file, similar to what was done in core recently.

https://github.com/owncloud/core/pull/41607 has a whole lot of code that got the acceptance tests infrastructure running in CI in core. One of the things done was to convert the Behat test config from YML to PHP. That now also needs to be done in the app repos, because the core `tests/acceptance/run.sh` now expects the Behat config to be in a PHP file.


phil-davis:
I will need to do similar to this in each ownCloud classic app repo.

ghost (approved):

